PR #5, opened 2020-11-18
Add persistent socket logic from elfman03.

jasonacox:
* Add a persistent socket that is not closed on every operation
* Add a respawing of socket if there is a socket error during send or receive (attempt to handle windows error 10054)
* Add a few "scenes" supported by color bulbs

### DIFF
--- a/tinytuya/__init__.py
+++ b/tinytuya/__init__.py
@@ -17,12 +17,14 @@
         dev_type (str): Device type for payload options (see below)
 
  Functions 
-    json = status()          # returns json payload
-    set_version(version)     #  3.1 [default] or 3.3
-    set_dpsUsed(dpsUsed)     # set data points (DPs)
-    set_retry(retry=True)    # retry if response payload is truncated
-    set_status(on, switch=1) # Set status of the device to 'on' or 'off' (bool)
-    set_value(index, value)  # Set int value of any index.
+    json = status()                    # returns json payload
+    set_version(version)               # 3.1 [default] or 3.3
+    set_socketPersistent(False/True)   # False [default] or True
+    set_socketNODELAY(False/True)      # False or True [default]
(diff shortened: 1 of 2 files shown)


tinytuya/__init__.py
```diff
@@ -263,28 +263,55 @@ def __init__(self, dev_id, address, local_key="", dev_type="default", connection
...
     def __repr__(self):
         return '%r' % ((self.id, self.address),)  # FIXME can do better than this
 
+    def _get_socket(self,renew):
+        if(renew and self.socket!=None):
+            self.socket.close()
+            self.socket=None
+        if(self.socket==None):
+          self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+          #s.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
```
elfman03, line 282: I'm not sure if some devices need NODELAY.  
I think I got fewer 10054 after commenting it out.  
Let me know if you want me to make a branch without this change.

tinytuya/__init__.py
```diff
@@ -263,28 +263,55 @@ def __init__(self, dev_id, address, local_key="", dev_type="default", connection
...
+            data = self.socket.recv(1024)
+            # Some devices fail to send full payload in first response
+            if self.retry and len(data) < 40:  
+                time.sleep(0.1)
+                data = self.socket.recv(1024)  # try again
+            success=True
+            #s.close()
+          except:
+            print("exception with low level tinytua socket!!!.  will retry!!!")
             time.sleep(0.1)
```
elfman03, line 312: This should probably be a proper log message instead of print.  For debugging I like the print but let me know if you want me to make a branch with a log instead.

elfman03 (commented):
Pointing out a few things in my code that you might want changed.

Issue #2, opened 2020-10-25
[WinError 10054]


burakmartin:
Sometimes this error appears, probably because a connection hasnt been closed before connecting to it again. How would I make sure that a connection to a device is closed?

jasonacox:
Interesting. The `_send_receive()` function calls `s.close()` each time.

What is your `python --version`?  Also, what version of Windows are you using?

elfman03:
I see this too.

Windows 10 (19042.572)
python --version : 3.8.6

`C:\Users\Perez family\Desktop\projector_control>python tuya_test.py
on
Traceback (most recent call last):
  File "tuya_test.py", line 23, in <module>
    data=tuya.turn_on()
  File "C:\Users\Perez family\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0\LocalCache\local-packages\Python38\site-packages\tinytuya\__init__.py", line 449, in turn_on
    self.set_status(True, switch)
  File "C:\Users\Perez family\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0\LocalCache\local-packages\Python38\site-packages\tinytuya\__init__.py", line 423, in set_status
    data = self._send_receive(payload)
  File "C:\Users\Perez family\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0\LocalCache\local-packages\Python38\site-packages\tinytuya\__init__.py", line 280, in _send_receive
    s.send(payload)
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
`

elfman03:
Sample app.  If I run it back to back, sometimes it runs, sometimes it gives the error a few times then runs again.

tuya=tinytuya.BulbDevice('bulb_id_here','ip_addr_here','key_here')
tuya.set_version(3.3)
data=tuya.status()
#print('Dictionary %r' % data)
if(data['dps']['1']==False):
  print("off")
else:
  print("on")
data=tuya.turn_on()
print('PostDictionary %r' % data)
data=tuya.set_colour(0xde,0xad,0)
print('PostDictionary %r' % data)
data=tuya.status()
#print('Dictionary %r' % data)
if(data['dps']['1']==False):
  print("off")
else:
  print("on")

elfman03:
Another sample:

[...]

PR #7, opened 2020-11-24
add set_socketRetryLimit(integer)

elfman03:
More for #2.  Extends #6 

Adds
set_socketRetryLimit(integer)      # retry count limit [default 5]

If retry limit is exceeded, the exception will be rethrown up to the client app.

to remove all retries set retry limit to 0 (or a negative number).  Will then see 10054 error on susceptible systems.

Error generating case for me:

```python
tuya.set_version(3.3)
#tuya.set_socketNODELAY(False)
#tuya.set_socketPersistent(True)
tuya.set_socketRetryLimit(0)
for i in range(2000):
    data=tuya.status()
    if(data['dps']['1']==False):
        print(str(i)+": off")
    else:
        print(str(i)+": on")
```

Working case for me is with the default 5 (I haven't seen more than like 2 retries under limited testing)

```python
tuya.set_version(3.3)
tuya.set_socketNODELAY(False)
tuya.set_socketPersistent(True)
#tuya.set_socketRetryLimit(0)
for i in range(2000):
    data=tuya.status()
    if(data['dps']['1']==False):
        print(str(i)+": off")
    else:
        print(str(i)+": on")
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Classes
     set_version(version)               # 3.1 [default] or 3.3
     set_socketPersistent(False/True)   # False [default] or True
     set_socketNODELAY(False/True)      # False or True [default]
+    set_socketRetryLimit(integer)      # retry count limit [default 5]
     set_dpsUsed(dpsUsed)               # set data points (DPs)
     set_retry(retry=True)              # retry if response payload is truncated
     set_status(on, switch=1)           # Set status of the device to 'on' or 'off' (bool)
--- a/tinytuya/__init__.py
+++ b/tinytuya/__init__.py
@@ -21,6 +21,7 @@
     set_version(version)               # 3.1 [default] or 3.3
     set_socketPersistent(False/True)   # False [default] or True
     set_socketNODELAY(False/True)      # False or True [default]
+    set_socketRetryLimit(integer)      # retry count limit [default 5]
     set_dpsUsed(dpsUsed)               # set data points (DPs)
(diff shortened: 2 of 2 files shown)


jasonacox:
Thanks @elfman03  - Merged.

arrmo:
Seeing this helping some errors I had as well, thanks! BTW, is this a limitation on the Tuya side (i.e. working around their responses)? Just curious 😄.

jasonacox:
Thanks @arrmo - I'm curious, to help with the errors are you using these new options or just the defaults in the 1.0.5 version? 

```python
tuya.set_socketNODELAY(False) #default is True
tuya.set_socketPersistent(True)  #default is False
```

To be fair to Tuya, the Tuya based devices are really designed to be monitored and managed via the TuyaCloud.  For those of us who want to control them directly / locally, you need to decipher the protocol or use a tool like **tinytuya** to help. ;-)

elfman03:
FYI, for me the retry=5 default made the biggest difference in preventing death. The persistent socket furthur dropped the number of times a retry is needed down but the retry was working ok even without the persistence.  The NODELAY, I'm not sure it helps one way or the other.  I am also curious what other people see with the persistence option.

arrmo:
> are you using these new options or just the defaults in the 1.0.5 version?

Just using the defaults ... OK?

> For those of us who want to control them directly / locally, you need to decipher the protocol or use a tool like **tinytuya** to help. ;-)

Agreed! And thanks for this tool - much appreciated!

PR #33, opened 2021-02-16
Bugfix/set white percentage colour temp computation fix

osomdev:
Bugfix for the colour temperature computation in set_white_percentage.


jasonacox:
Good catch! Thank you, @osomdev

PR #45, opened 2021-05-18
Expose server error to user

eykamp:
Seems nicer than crashing!


jasonacox:
Thanks @eykamp - Great addition!

PR #46, opened 2021-05-18
Emit error rather than crash

eykamp:
In my situation, I am getting "Error" in my status, but not "dps".  The status packet looks a lot like the one generated by error_json, so returning it seems logical.  The only caveat here is that I cannot get my device to work, so I am not certain that there is no Error in a valid status message, though it seems unlikely.  If it is there, you could add a check to ensure "dps" is missing, but I suspect that will be unnecessary.

### DIFF
--- a/tinytuya/__init__.py
+++ b/tinytuya/__init__.py
@@ -1551,6 +1551,9 @@ class BulbDevice(Device):
         if not status:
             return error_json(ERR_JSON,"state: empty response")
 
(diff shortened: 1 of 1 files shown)


jasonacox:
@eykamp Thanks for this!  Yes, I can see how this would prevent a crash.  I will accept this but may also update it to be more like this:

```python
        if self.DPS not in status.keys():
            return status
```

PR #47, opened 2021-05-18
Update terminology

eykamp:
I found the use ID/Product ID confusing and not consistent with other usages.  Maybe this is better.

### DIFF
--- a/tinytuya/__init__.py
+++ b/tinytuya/__init__.py
@@ -1765,10 +1765,10 @@ def deviceScan(verbose=False, maxretry=MAXCOUNT, color=True, poll=True):
                     pass
             if(verbose):
                 if(dname == ""):    
-                    print("%s%s Device Found%s [%s payload]: %s%s%s\n    ID = %s, Product ID = %s, Version = %s" % (
+                    print("%s%s Device Found%s [%s payload]: %s%s%s\n    Device ID = %s, Device Key = %s, Version = %s" % (
                     normal, version, dim, note, subbold, ip, dim, gwId, productKey, version))
                 else:
-                    print("%s%s%s [%s payload]: %s%s%s\n    ID = %s, Product ID = %s, Version = %s" % (
+                    print("%s%s%s [%s payload]: %s%s%s\n    Device ID = %s, Device Key = %s, Version = %s" % (
                     normal, dname, dim, note, subbold, ip, dim, gwId, productKey, version))
             try:
                 if poll:


jasonacox:
@eykamp  I see you want to change "Product ID" to "Device Key" in the scan function.  My hesitation to accept this change is that most people will believe that this is the "Local Key" for the device used for encryption, but that is not the case.  This identifier is called "Device Key" by Tuya, but it is not the encryption key.  

I chose "Product ID" because it seems to be a better description.  If you have 3 of the same type of plugs, for example, they will all show the same Device Key.  That is why I chose to use "Product ID".  Let me know what you are thinking.  I can be convinced otherwise, but want to help remove confusion of people thinking the `scan` results provide a local key instead of using the `wizard` to get the local key.

One thought, for cases where we have the real local key (dkey), we could display that in addition to the product ID.

eykamp:
You are probably right; I have been unable to get my device to connect to the tinytuya library (which is why I had so many suggestions about reporting errors) and part of the problem could be confusion on my part about what is what.

It sounds like Product ID is an appropriate name. Am I correct and understanding that the wizard is the authoritative source of the encryption key needed to connect to the device?  If so, perhaps a better solution would be to discard my pull request and add some text to the scan function alerting users that they need to get the key from the wizard.

I have not been able to get the wizard to run successfully because something is not registered on the tuya API page.

I'll try to sort that out tomorrow night, and maybe that will clarify things.

Thanks,

Chris

eykamp:
I got my device working, and have a better insight into why I was confused.  You should not merge the pull request.

But I do think restructuring the scan output would be helpful.  I would suggest moving the Product ID up to the upper line and the IP address down to put all the info the program needs on one line, and the metadata on another.  I would also add the local key needed to connect to the device.

i.e. convert this:

```
Light Bulb [Valid payload]: 192.168.0.5
    Product ID = 53601845481edac268f2, Device Key = key5vf3gy2wtuuie, Version = 3.3
```

to this:

```
Light Bulb   Product ID = key5vf3gy2wtuuie  [Valid payload]: 
    Address = 192.168.0.5,  Device ID = 53601845481edac268f2,  Local Key = c35e4e1320cbbb46,  Version = 3.3
```

That still doesn't feel quite right, but it does seem to better organize the information in a way that is most useful (or at least would have been to me).  It places the data the Device constructors need together, and flows better from the docs that say "Controlling and monitoring Tuya devices on your network requires the following: IP, ID, Version, Local_Key"  (though I might suggest renaming IP to Address in this sentence to better match the way you refer to ip addrs in the code, and maybe replacing ID with Device ID).

Also, when poking around on the Tuya API site, I noted that there also appear to be more regions now; there is an Eastern US, and a Western Europe, for example.

jasonacox:
I will close this without a merge, but you have a great idea and I do think that would be helpful.  I'm not where I can do much with any updates this week, so if you have time and interest, please feel free to submit another PR.  Otherwise, I'll add it to my backlog.  Thanks for the feedback and help!

PR #51, opened 2021-06-03
repair print on devices.py

SmookeyDev:


jasonacox:
@SmookeyDev Thanks for the fix!

PR #54, opened 2021-06-09
Add bulb type C corresponding to Feit Electric Dimmers sold at Costco

davidlie:
These Feit Electric Dimmers use a slightly different DPS encoding than the ones already supported.  Rather than try to automatically detect which bulb, I added an API to allow explicit setting of the bulb type and added a new bulb type to support the Feit Dimmers.

### DIFF
--- a/tinytuya/__init__.py
+++ b/tinytuya/__init__.py
@@ -705,6 +705,9 @@ class XenonDevice(object):
         else:
             self.dps_to_request.update({str(index): None for index in dp_indicies})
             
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks for the PR David.  I like the idea of being able to alter the Bulb type manually and appreciate the 'Type C' addition.  I'll update the README.

davidlie:
Your welcom and happy to help.  Thanks for sharing this project, it certainly saved me a ton of work!

PR #98, opened 2022-01-13
Displays raw data better.

mores:
Some of the debug output is getting converted to actual characters. 

Before Fix:
`DEBUG:payload generated=b'\x00\x00U\xaa\x00`

After Fix:
`DEBUG:payload generated=b'000055aa00`

### DIFF
--- a/tinytuya/__init__.py
+++ b/tinytuya/__init__.py
@@ -553,7 +553,7 @@ class XenonDevice(object):
                         time.sleep(0.1)
                         data = self.socket.recv(1024)  # try to fetch new payload
                     success = True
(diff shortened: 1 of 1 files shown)


mores:
Or if you prefer this output:

`DEBUG:payload generated=b'00 00 55 aa 00 00`

Then modify code to add seperator:
`log.debug("received data=%r", binascii.hexlify(data,' '))`

jasonacox:
Originally the debug output was designed so you could copy the payload and lift it into a python session to troubleshoot.  Your PR would still work, would just require `binascii.unhexlify().  I agree this is more readable especially with the space.  Thanks for the PR!

PR #115, opened 2022-03-20
Add galaxy lamp example

knrdl:
Hi,

my girlfriend recently bought the Galaxy Projector from galaxylamps.co. It's basically a Tuya RGB Bulb with some neat extras. I would like to share the code to control this device, maybe it is helpful for somebody.

Your code and instructions made it possible for me to control the device without the app. Thanks a lot!


jasonacox:
Nice!  Very clean code @knrdl - thanks for the PR!

PR #120, opened 2022-03-28
Added support to get the MAC address and local IP address in the Wizard

frodeheg:
IP address and MAC address support as promised.


jasonacox:
Thanks @frodeheg  - what PyPI modules (pip install) did you use for these new imports?

frodeheg:
pip install netifaces
pip install python-nmap
pip install getmac

PR #127, opened 2022-04-25
Support optional version parameter

dominikkarall:
Some APIs are already available in a higher version. This change supports an optional argument `version` which is "1.0" by default.


jasonacox:
@dominikkarall thank you!  I merged your PR even though there is a _tuyaplatform() function since it may not be obvious to everyone else.  My long term plan is to have wizard use the tinytuya Cloud class functions instead.

PR #135, opened 2022-05-23
Some cleanup from pylint

mafrosis:
Hi, I use `pylint` a lot in my projects, and found a lot of low hanging fruit in the tinytuya codebase which I fixed up this morning.

A lot of this stuff is a matter of opinion, so feel free to close this PR with maximum prejudice 🥲 I also separated into different commits, so you could review each one and rebase out the ones you don't agree with 🤷 

* Remove trailing whitespace
* Pythonic imports
* Lazy logging calls
* Pythonic parentheses on if/return statements
* Fixed some indentation

I also included a `.pylintrc` which can be modified to taste - I use a lot of `pylint` ignores, to filter out the rules which are imo stylistic as opposed to objectively good practice:

> too-many-lines, line-too-long, too-many-ancestors, too-many-arguments, too-many-branches, too-many-instance-attributes, too-many-locals, too-many-nested-blocks, too-many-return-statements, too-many-statements, too-few-public-methods, ungrouped-imports

If you like where this is going, the next pass would be on the more nefarious pylint warnings (not all of which should be "fixed", of course):

* `bare-except`/`broad-except` which encourage good exception handling practice; hard to fix as they can change application behaviour
* `unspecified-encoding` which exists to help ensure files are written with consistent encoding on different platforms
* `redefined-outer-name`/`redefined-builtin` where a variable is clobbering an existing name; a can cause unexpected bugs

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,5 @@
+[MESSAGES CONTROL]
+disable=bad-whitespace, consider-iterating-dictionary, import-error, invalid-name, missing-docstring, no-else-return, no-member, too-many-lines, line-too-long, too-many-ancestors, too-many-arguments, too-many-branches, too-many-instance-attributes, too-many-locals, too-many-nested-blocks, too-many-return-statements, too-many-statements, too-few-public-methods, ungrouped-imports, use-dict-literal
+
+[SIMILARITIES]
+min-similarity-lines=8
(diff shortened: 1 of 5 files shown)


jasonacox:
Accepted!  Thank you @mafrosis!  I appreciate the clean up. These look good.

 The only style complaint I had was the superfluous params.  I find `not (a <= b <= c)` is easier to grok (human parser) than `not a <= b <= c` but not a pylint hill to die on.  😀

mafrosis:
FWIW I agree with you on the round brackets 😄 

You can always bring those back in future and add `superfluous-parens` to the ignore line in `.pylintrc`

Issue #74, opened 2021-10-24
IR REMOTE CONTROL


adlerlinhares:
HI, first of all apologize my poor english.
Im facing a major issue trying to code the IR remote control.
Is there anyway to find out how to control it with tinytuya?

Im using yout code:

```python
import tinytuya

d = tinytuya.OutletDevice('XXX, "YYY", "ZZZ")
d.set_version(3.3)
d.set_socketPersistent(True)
payload = d.generate_payload
print(" > Send Request for Status < ")
payload = d.generate_payload(tinytuya.DP_QUERY)

print(" > Begin Monitor Loop <")
while(True):
    # See if any data is available
    data = d.receive()
    print('Received Payload: %r' % data)

    # Send keyalive heartbeat
    print(" > Send Heartbeat Ping < ")
    payload = d.generate_payload(tinytuya.HEART_BEAT)
    d.send(payload)

    print(" > Send Request for Status < ")
    payload = d.generate_payload(tinytuya.DP_QUERY)
    d.send(payload)
    
    
    # NOTE Some smart plugs require an UPDATEDPS command to update power data
    print(" > Send DPS Update Request < ")
    payload = d.generate_payload(tinytuya.UPDATEDPS)
    d.send(payload) 
```

AS result i get the follow message:

`Received Payload: {'dps': {'201': '{"control":"study_exit"}'}}`

Is there a way that i can have a list of commands?
what shoud be this "study_exit"?

Thank you so much!
Best regards

ClusterM:
Seems like the `CONTROL_NEW` command uses some alternative encryption key/method. We need to decompile the Smart Life app to understand it.

ClusterM:
I still can't control it locally but it works fine using the official Tuya Cloud API. It's offtopic but maybe it will help somebody. There is a method to send RAW IR signals: https://developer.tuya.com/en/docs/cloud/6c376807b5?id=Kb3oeb91u35ga
`remote_id` must be any remote ID (so you need to create at least one, I have no idea why it's required) and `code` is a hexadecimal string of 16-bit (little-endian) values with signal lengths and gap lengths in microseconds. E.g. code `112233445566778899aa` means:
* `1122` = 8721 microseconds of signal
* `3344` = 17459 microseconds of gap
* `5566` = 26197 microseconds of signal
* `7788` = 34935 microseconds of gap
* `99aa` = 43673 microseconds of signal
And so on.
[...]

PR #137, opened 2022-06-01
Add Universal IR Controller DP instructions

mont5piques:
Add a detailled procedure to control Universal IR Controllers using JSON DPs
Related to #74


jasonacox:
Thanks @mont5piques !

PR #140, opened 2022-06-14
Apply new socket options to existing open sockets

uzlonewolf:
I'm using the socket timeout during receive() as a crude timer and need to change it on the fly, but set_socketTimeout() was not applying to existing open persistent sockets.  This update makes set_socketTimeout() and set_socketNODELAY() update immediately, and causes set_socketPersistent() to close the socket if persistent is changed to False and a socket is open.


jasonacox:
Great fix!  Thanks for this, @uzlonewolf !

jasonacox:
Updated code released in [v1.6.0](https://github.com/jasonacox/tinytuya/releases/tag/v1.6.0)

PR #142, opened 2022-06-15
Break the Outlet/Cover/Bulb/Cloud modules out into separate files

uzlonewolf:
This separates the OutletDevice, CoverDevice, BulbDevice, and Cloud classes out into separate files.

Effectively __init__.py was renamed to core.py, the above classes were moved into new files, and a new __init__.py that imports all of the above was created.  No code was changed, though various `import ...` statements were changed to ensure everything a module needed was imported.

I'm sure some of the comments/documentation will need to be updated.


jasonacox:
I have always wanted to do this but never had time!  You made my day! 😄  Thank you @uzlonewolf! 

I'll go through all the code and comments and make the necessary updates.

PR #144, opened 2022-06-17
fix logging calls in XenonDevice.detect_available_dps

pkasprzyk:
log calls use `self` instead of `log`, which crashes with `AttributeError`


jasonacox:
Great catch!  Thanks for the fix, @pkasprzyk !

PR #146, opened 2022-07-02
Create 'Contrib' and add ThermostatDevice

uzlonewolf:
Also has an example under Contrib/examples/

I tried to make it self-contained so you only need to `from tinytuya import Contrib` without needing to also `import tinytuya`


jasonacox:
Thanks @uzlonewolf!  I'm very happy to have this first contributed device expansion.  Merged.

Discussion thread: https://github.com/jasonacox/tinytuya/discussions/139

PR #147, opened 2022-07-02
Add Cloud API syncing with auto-retry

### LINKED ISSUE #143: [Server] Feature Request: Add a way to sync or put keys for devices
Hello,

i try to use the API server ( i want to use it as an add-on in HA and an integration to get the devices into HA without the cloud). Currently it finds my devices just fine and i could get its status but the key is missing and i get an error. I also don't see a way to set the key of a device or the API server taking a Tuya IoT Token for querying itself against the cloud to get the local key. 

It would be nice to have either a endpoint which can set the local key or having the API Server ask Tuya itself. For me it would be nice if maybe it can support both and only activate the Tuya API scraping only if credentials are given. 

If you want i can implement, test and PR it. Just give shoot me a info if its ok.

Thanks

uzlonewolf:
Closes #143

Loading of the API keys needs improvement.  Possibly via another json file or environmental variables?

This pull:
*) Adds a `/sync` URL to force a resync
*) Adds a `/cloudconfig/{apiKey}/{apiSecret}/{apiRegion}/{apiDeviceID}` URL to load the Cloud API login info
*) When a new device is detected which is not in the known-device list, it resyncs from the Cloud API
*) If said device is still not found, it retries 5 times waiting 30 seconds between tries
*) If a manual resync is requested, it again looks at all the new devices (but does not retry)
*) After resyncing, it saves devices.json if SAVEDEVICEFILE is set to True
*) It eliminates the `havekeys` variable in favor of `len(tuyadevices) > 0`
*) For the UDP listeners, it changes the socket timeout to 5 seconds as I was having issues with the threads hanging on `KeyboardInterrupt`
*) Exit on `SystemExit` as well as `KeyboardInterrupt`
*) In index.html, if a device does not have a name then use the ID instead.  Otherwise the URL is not clickable
*) Update Cloud.py to include the MAC if it's available


jasonacox:
Great additions and fixes! Thanks @uzlonewolf - updates merged.

PR #151, opened 2022-07-05
Update Cloud.py

Paxy:
Added getconnectstatus(deviceId) function to check is device connected to Tuya cloud


jasonacox:
Great addition!  Thanks @Paxy !

PR #153, opened 2022-07-06
Break socket recv() out into its own function

### LINKED ISSUE #152: Multiple received packets results in none getting decoded
While debugging something completely unrelated I noticed a `DEBUG:tinytuya.core:incomplete payload=...` go flying by.  Further investigation reveals the "incomplete" payload actually contained 3 completely separate payloads: one which was just requested, and 2 async DPS updates.  This "multiple = none" explains why I was sometimes only seeing one or 2 of my thermostat sensor list updates instead of all 5.

I see there is a "length" field in the header; is there any reason this is not used?  `unpack_message()` blindly assumes everything minus the footer length is part of a single packet.

As for fixing this, I can think of 2 options: 1) change the `self.socket.recv(1024)` to only read the header length and then do a 2nd recv() for the body once we know the length (leaving any remaining message(s) for the next `receive()` call to deal with), or 2) keep the large read and return it in the dict as "next_packet" or something and make the user handle the detection of this and the feeding of it back into `receive()`.  I'm really not feeling option 2 as in addition to being clunky you are also going to have problems if a message isn't completely received in a single `recv()` call.

uzlonewolf:
This PR:

1. Moves `self.socket.recv()` into its own function
2. Makes the recv() only read the header and payload of a single message
3. Parses the message into the `TuyaMessage()` named tuple
4. Moves the retcode into its own set of variables as it is not actually part of the header
5. Spits out `log.debug()` messages if the header, footer, or CRC are wrong
6. Misc other debugging message tweaks

Closes #152

### DIFF
--- a/tinytuya/core.py
+++ b/tinytuya/core.py
@@ -161,14 +161,15 @@ LAN_SET_GW_CHANNEL = 252
 PROTOCOL_VERSION_BYTES_31 = b"3.1"
 PROTOCOL_VERSION_BYTES_33 = b"3.3"
 PROTOCOL_33_HEADER = PROTOCOL_VERSION_BYTES_33 + 12 * b"\x00"
-MESSAGE_HEADER_FMT = ">4I"  # 4*uint32: prefix, seqno, cmd, length
-MESSAGE_RECV_HEADER_FMT = ">5I"  # 4*uint32: prefix, seqno, cmd, length, retcode
+MESSAGE_HEADER_FMT = ">4I"  # 4*uint32: prefix, seqno, cmd, length [, retcode]
+MESSAGE_RETCODE_FMT = ">I"  # retcode for received messages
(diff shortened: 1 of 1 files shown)


jasonacox:
Nice!  Thanks @uzlonewolf  - This is a very clean / elegant approach and I'm trying to remember why we/I didn't address this already.  I remember having it in my TODO list to address the multi-palyload condition.  

On the second bit, I understand and appreciate the exception approach.  I can see that as a breaking change for some users.  I want to run through some tests to see how serious that would be before decide on version increment, etc.  

Thank you!

uzlonewolf:
It shouldn't be a breaking change as it's only called from within the try/except block in `_send_receive(()` - the `except Exception as err:` line will catch it and retry if there are any retries remaining.

Though now that I think about it, that retry may be a breaking change - previously, decode errors were caught a bit further down and resulted in an error being returned instead of a retry.  To restore this response we would just need to throw a specific exception and have a catch for it.

PR #154, opened 2022-07-06
Do not retry on decode error (restores previous behavior)

uzlonewolf:
As per comment in #153

### DIFF
--- a/tinytuya/core.py
+++ b/tinytuya/core.py
@@ -209,6 +209,9 @@ error_codes = {
     None: "Unknown Error",
 }
 
+class DecodeError(Exception):
+    pass
+
 # Cryptography Helpers
 class AESCipher(object):
     def __init__(self, key):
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks @uzlonewolf - Honestly I was still trying to determine if the new behavior might be better.  But this is probably better for more of an incremental rollout (v1.6.2 vs 1.7.0).

With the TuyaHeader enhancement are you seeing better reliability with your thermostats?  Are there cases where the retry would be helping?  I have a few smartbulbs that are generally problematic (dirty payloads) so I'm running a battery of tests against them.

uzlonewolf:
I *just* finished this so I don't have any real data yet.  With the number of "incomplete payloads" I was getting due to the thermostat firing off 5+ updates back-to-back I would be surprised if it didn't help.

jasonacox:
Understood.  Thanks for the update!  Let me know how it goes.

uzlonewolf:
So far it seems to be really solid, haven't noticed any errors or missed commands.  I do still get a few "incomplete payloads" every now and then, but they are all null packets (no payload).  It seems `_decode_payload()` gets mad when it is fed an empty string (`b''`).  Not sure how much I care about that though as the try/except just results in returning `None`.

The more I think about it the more I think read retries should be separated from send retries.  If either a null packet or a corrupted packet is received it should retry the read a few times and only retry the send if we get a timeout.

uzlonewolf:
I'd be interested in knowing if the "search for start prefix" addition in #158 helps with those problematic smartbulbs.

jasonacox:
I love it!  I'm going to merge and fire off some tests tonight to see what happens.  Thanks, @uzlonewolf

uzlonewolf:
I'm definitely interested in how those smartbulbs react.  If they're still giving you trouble can you send me more information on them?  I always love a good challenge :grin:

jasonacox:
Honestly, @uzlonewolf, I can't get it to fail now!  You are a wizard! 🪄 

I have a few more tests that I'm running tonight. I'll let you know if I see a problem.

uzlonewolf:
I'm glad it's finally cooperating :sunglasses:  Changing send() to sendall() was pretty much the only thing left to try.  I still may do that just to be complete :smiley:  I've seen some stuff come flying off the TCP stack over the years so I've gotten pretty good at dealing with it.

PR #155, opened 2022-07-06
More verbose debug logging on decode error

uzlonewolf:
Right after that last PR #154 I realized I forgot about debug logging.


jasonacox:
Or I'm just to quick on the merge button!  😆   Good catch.  Thanks again for your PRs.

PR #157, opened 2022-07-06
Add schedule editing to Thermostat and various fixes

uzlonewolf:
It still needs some work as currently the schedule temperatures must be in degrees C and the time is minutes since midnight, but I wanted to get something as well as the other fixes in before 1.6.2 drops.

This PR contains lots of documentation updates, reworks the temperature sensor list to be more iterable and adds accessing as a dict, fixes the temperature sensor list 2-5 parsing/updating, and adds an initial schedule editor.


jasonacox:
Thanks @uzlonewolf - I plan to push out 1.6.2 in a few hours.

jasonacox:
Live at https://pypi.org/project/tinytuya/1.6.2/

PR #163, opened 2022-07-29
Create TuyaSmartPlug-example.py

fajarmnrozaki:
Proposing the example of python script file for monitoring the Tuya Smart Plug (1 Pack) that enables to sense a voltage, a current, a power of a connected device with its smart plug.


jasonacox:
Thanks for the example script @fajarmnrozaki !  I have accepted the PR.  I notice there is device ID and key information in the script.  I assume these are bogus but will redact them just in case.

fajarmnrozaki:
Thank you so much @jasonacox for merging the pull request in the main branch, and thanks for correcting the mistakes that I put my own ID in the example script of TuyaSmartPlug-example.py

PR #165, opened 2022-08-01
Fixing the minor mistake

fajarmnrozaki:
In line 65th, the %S in timer format strftime was not written. Actually, it is no big deal, but it will affect the query time inside MySql for the second unit precisely.


jasonacox:
Thanks for the fix, @fajarmnrozaki !

PR #166, opened 2022-08-07
Rework payload_dict and realign the command list to match Tuya's API

uzlonewolf:
When I started throwing random commands at my thermostat to see what it responded to, I noticed DP_QUERY_NEW was defined as 16 but payload_dict stored it as hex byte 0f (which would be 15).  The more I dug into `payload_dict` and `generate_payload()` the less I liked about how it worked, so I rewrote it ^_^ .  Now it functions more like an override - if the command is not explicitly overridden (such as with device22's DP_QUERY) it simply uses the command as passed to `generate_payload()`, and if the json dict key "command" is not defined then it just sends everything ("command" is set to `"command": {"gwId": "", "devId": "", "uid": "", "t": ""}`).  Technically the `AP_CONFIG` and `DP_QUERY` entries can be deleted from `payload_dict` completely, but for now I left them.

I also noticed the list of commands didn't really make sense.  It looked like it was copy-n-pasted from the list of MCU serial commands, however those have nothing to do with the local API / TCP socket commands.  So, I used the LAN protocol defines from https://github.com/tuya/tuya-iotos-embeded-sdk-wifi-ble-bk7231n/blob/master/sdk/include/lan_protocol.h and removed the invalid ones.  I also added the "official" names to the comments and did some text aligning.

There are also 2 unrelated 1-liners I threw in: socket.send() is now socket.sendall(), and a copy-n-paste error in the IR Remote example was corrected.

### DIFF
--- a/server/tinytuya
+++ b/server/tinytuya
@@ -1,0 +1,1 @@
+../tinytuya
--- a/examples/Contrib/IRRemoteControlDevice-example.py
+++ b/examples/Contrib/IRRemoteControlDevice-example.py
@@ -1,5 +1,5 @@
 
-# TinyTuya ThermostatDevice Example
+# TinyTuya IRRemoteControlDevice- Example
 # -*- coding: utf-8 -*-
 """
  Example script using the community-contributed Python module for Tuya WiFi smart universal remote control simulators
--- a/examples/Contrib/IRRemoteControlDevice-example.py
+++ b/examples/Contrib/IRRemoteControlDevice-example.py
@@ -1,5 +1,5 @@
 
-# TinyTuya IRRemoteControlDevice- Example
+# TinyTuya IRRemoteControlDevice Example
 # -*- coding: utf-8 -*-
 """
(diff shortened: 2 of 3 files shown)


jasonacox:
I love this! All of these make sense and improve the project. Brilliant work!  I don't see a risk of breaking the API so will release this as as a minor update, `v1.6.5`.

Thanks @uzlonewolf !!!

uzlonewolf:
I'm not entirely sure how it slipped in, but it seems I inadvertently added a symbolic link to `../tinytuya` under `server/`.  Not sure if you want to keep it or remove it.

server/tinytuya
```diff
@@ -0,0 +1 @@
+../tinytuya
```
uzlonewolf, line 1: Oops, this wasn't supposed to be committed...

jasonacox: I've done the same, many times.  Easy fix!  Updated.

PR #167, opened 2022-08-09
Added SocketDevice

Felix-Pi:
Added SocketDevice to receive power consumption data

### DIFF
--- a/tinytuya/Contrib/README.md
+++ b/tinytuya/Contrib/README.md
@@ -34,6 +34,21 @@ In addition to the built-in `OutletDevice`, `BulbDevice` and `CoverDevice` devic
     ir.send_button(button)
     ```
 
+### SocketDevice
+
+* SocketDevice - A community-contributed Python module to add support for Tuya WiFi smart sockets
+* Author: [Felix Pieschka](https://github.com/Felix-Pi)
+
+    ```python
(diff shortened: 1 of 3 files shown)


jasonacox:
Brilliant! Thanks @Felix-Pi !  Great work on the extension.  

I merged this but suggest you consider adding a few things:

* There are 3.1 and 3.3 devices that use DPS 4, 5 and 6 for current, power and voltage (see [here](https://github.com/jasonacox/tinytuya#version-31---plug-or-switch-type)).  I have some of these. You could add some basic logic to check for DPS 18 and if it isn't there, use 4, 5 and 6.
* Error handling for non-energy plugs - If a device does not show up with the needed DPS values, instead of allowing a `KeyError` exception, catch it and respond back with an error indicating the device does not appear to have energy measurement (see the `error_json()` [function](https://github.com/jasonacox/tinytuya/blob/cfc4aa788ee0aff32f9abd47d84980c6e023e245/tinytuya/core.py#L354) if you want to use it for a response).

Felix-Pi:
Thank you for your input! I'll look into it as soon as possible 👍🏼

jasonacox:
@Felix-Pi your code is now live in [v1.6.6.](https://pypi.org/project/tinytuya/1.6.6/).

Thanks again for your help!

PR #169, opened 2022-08-15
skip DPS detection if it has already been set

pawel-szopinski:
since detect_available_dps() function takes considerable amount of time to run, skip it if DPS has already been set for a device with set_dpsUsed() function


jasonacox:
Great addition! Thanks for the PR.  Merged!

PR #174, opened 2022-08-31
ThermostatDevice update

uzlonewolf:
This adds automatic C/F detection and conversion when changing schedules, and updates the sensor temperature attributes for the new firmware version that was recently pushed out.


jasonacox:
Thanks @uzlonewolf !

Issue #148, opened 2022-07-04
3.4 protocol is not supported


koyaya:
Access rejected by 192.168.0.138: Unexpected Payload from Device

jasonacox:
Hi @koyaya - thanks for opening this issue.  You are correct.  TinyTuya does not yet support 3.4 protocol.  I don't have a 3.4 device yet so I'm limited on what I can do.  I'm very happy to accept any PRs to help us expand support to 3.4.

Can you provide a bit more context to help us?

* What device are you using (a link would be helpfpul)?
* Have you find another project that works with your 3.4 device?

Ref: 
- https://github.com/codetheweb/tuyapi/issues/481
- https://github.com/rospogrigio/localtuya/issues/812

jasonacox:
Note on protocol: https://github.com/codetheweb/tuyapi/issues/481#issuecomment-921756711

@harryzz has figured out the protocol and added code to a forked version of tuyapi (node library) to support 3.4.  This could be incorporated into TinyTuya if someone has a device they can use to test.

koyaya:
i didn't find anything that supports 3.4
will be cool if you would be the first one to support it

i can help you

this is the device that i have

US $5.80  42%OFF | Tuya Smart Wifi Motor Switch Module 5V 12V 32V 220V RF
433 Radio Remote Control 4 Channels Inching Relay for Alexa Google Home
https://a.aliexpress.com/_vl0Ak2

harryzz:
Hi,
This gateway from silvercrest (lidl) support v 3.4:
https://paulbanks.org/projects/lidl-zigbee/#overview
available on ebay.de.

regards,
Zahari

koyaya:
what i have isn't a gateway, this is a smart relay, like a switch, only
with 4 channels

jasonacox:
Agree @koyaya , it would be great to get this added.  I will see if I can get one of these devices.

Hi @harryzz , thanks for your help!  Are you aware of any other smart plugs, lights or others that use 3.4?

pretoriano80:
@jasonacox i got this [Mini Smart Switch](https://a.aliexpress.com/_mOLnb2q) and it's a 3,4 device and it's also cheap.
With mine,on  **tinytuya scan** i get this:

 `Access rejected by 192.168.188.50: Unexpected Payload from Device`

jasonacox:
[...]

PR #182, opened 2022-09-12
v3.4 fixes for persistence and BulbDevice

uzlonewolf:
Brings non-persistence back for v3.4 devices and fixes BulbDevice.set_version() as reported in https://github.com/jasonacox/tinytuya/issues/148#issuecomment-1243300256

Edit: I ended up merging in the encrypting/encoding rewrite instead of making a new PR.
generate_message() now returns a namedtuple instead of encrypting and packing the message.  Any program which just passes its output to send() or _send_receive() will work fine, but any which either pack their own message or modify the output of generate_message() are gonna break.  _send_receive() now takes care of the packing and encrypting so it can re-encode whenever the socket is closed and reopened, and _get_socket() now takes care of negotiating the session key.

For non-persistent sockets, I noticed _send_receive() was closing and reopening the socket if it received a valid message with a null payload (an ack packet) which caused it to then miss the response it was looking for.  So, that has been updated to only close/re-open if the send fails or we do not receive an ack.

I also added a `version=` keyword to the device constructor so you can set it in one go instead of calling set_version() afterwards.

### DIFF
--- a/tinytuya/BulbDevice.py
+++ b/tinytuya/BulbDevice.py
@@ -229,7 +229,7 @@ class BulbDevice(Device):
             Type B has keys 20-29
             Type C is Feit type bulbs from costco
         """
-        self.version = version
+        super(BulbDevice, self).set_version(version)
 
         # Try to determine type of BulbDevice Type based on DPS indexes
         status = self.status()
--- a/tinytuya/BulbDevice.py
+++ b/tinytuya/BulbDevice.py
@@ -114,8 +114,8 @@ class BulbDevice(Device):
     has_colourtemp = False
     has_colour = False
 
-    def __init__(self, dev_id, address, local_key="", dev_type="default"):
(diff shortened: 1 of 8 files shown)


jasonacox:
Merged.  Thanks @uzlonewolf - you're incredible! Technically the API change with `_send_receive()` is allowed without breaking API contract since it is a "_" function, but I'll add this to the release notes.  

Also, I love the `version=` to constructor addition!  Old code will still work (as will anything using pytuya) but would allow us to update all the README and example scripts to a more intuitive (and if wanted, self descriptive) way.

```python
d = tinytuya.OutletDevice(
    dev_id='xxxxxxxxxxxxxxxxxxxxxxxx',
    address='x.x.x.x',
    local_key='xxxxxxxxxxxxxxxx',
    version=3.4)

print(d.status())
```

PR #186, opened 2022-10-04
Rework device finding for auto-IP detection, and unpack_message() retcode fix

uzlonewolf:
Update find() to use `select.select()` instead of a fixed wait so it processes broadcast packets as they come in.  It was also moved out of XenonDevice and renamed to find_device() so you can use it without creating a device.  It can also be used with an IP address instead of a device ID if you want to find out what device is at a certain IP (not sure why I added this one, this is getting into scanner territory).

I also noticed `unpack_message()` was returning a `list` (instead of the `int` it should be) if a packet retcode is processed, so that is now fixed.  I don't think anything in tinytuya ever uses the retcode, but my scanner uses it to help detect brute-force failures.


jasonacox:
Thanks @uzlonewolf !  I don't know why the commit was showing multiple core.py replacements, but I reviewed the final deltas and it looks good.  More importantly, it works GREAT!  Thank you! ❤️ 

```python
>>> tinytuya.find_device('0123456789abcdef0123')
('10.0.3.3', '3.3', '0123456789abcdef0123')

>>> tinytuya.find_device(address='10.0.3.3')
('10.0.3.3', '3.3', '0123456789abcdef0123')
```

```python
>>> d = tinytuya.OutletDevice('0123456789abcdef0123', 'Auto', '0123456789abcdef')
>>> d
('0123456789abcdef0123', '10.0.3.3')
>>> d.status()
{'devId': '0123456789abcdef0123', 'dps': {'1': False, '101': 0}}
```

uzlonewolf:
It's showing core.py as replaced because I used a different text editor for part of it and it converted all the line endings from \r\n to just \n without telling me, so I then had to commit it again with that fixed.

PR #187, opened 2022-10-04
Get local key from devices.json if not provided

uzlonewolf:
This adds a helper function to get the device data from devcies.json, and uses it to look up the local key if it is not provided.  It also makes the IP address argument optional (it will try to find the device if not provided) .  This means if you have a devices.json file containing the device key and the device is broadcasting on the local network, all you need to do is:

```python
d = tinytuya.OutletDevice( '0123456789abcdef0123' )
```

### DIFF
--- a/tinytuya/BulbDevice.py
+++ b/tinytuya/BulbDevice.py
@@ -114,7 +114,7 @@ class BulbDevice(Device):
     has_colourtemp = False
     has_colour = False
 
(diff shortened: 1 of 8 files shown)


jasonacox:
Thanks @uzlonewolf - you, my friend, are simply amazing! I thought about doing something similar to this multiple times but didn't find the motivation, mostly because the `find()` was so painfully slow. Thank you!

Issue #185, opened 2022-10-01
Add local dp_x handling based on known product ids


fhempy:
Hi,
what do you think about adding known mappings directly to the project as iobroker does?
See https://github.com/Apollon77/ioBroker.tuya/blob/master/lib/schema.json

They use the product id to identify the dp mappings. So you don't need any connection to the cloud if you know your localkey and product id.

I used it in my project as well where I use a combination of tinytuya and localtuya (because of asyncio) code:
https://github.com/fhempy/fhempy/tree/master/FHEM/bindings/python/fhempy/lib/tuya

Would be great to make one mapping list in tinytuya where everybody could contribute.

jasonacox:
I love the idea @fhempy !  Similar to the community driven [Contrib](https://github.com/jasonacox/tinytuya/tree/master/tinytuya/Contrib) extension for devices, we could add something like a "Products" extension. How would you like to see it work?  My first thought...

```python
from tinytuya import Products

product_id = 'MShdslm9Uw7Q59nN'

# Look up attributes for product
a = Products.attributes(product_id)
print("Device version = %d - name = %s - type = %s" % (a['version'], a['name'], a['type']))

# Look up data points for product
print("DPS:")
dps = Products.dps(product_id)
for point in dps:
    print("  %d = %s - %s", % (dps['id'], dps['name'], dps['type'])

# Advanced Device Connection
d = Products.connect(
       product_id,
       dev_id='DEVICE_ID_HERE',
       address='IP_ADDRESS_HERE',
       local_key='LOCAL_KEY_HERE')

d.turn_on()
if d.device() is 'led_switch':
    d.set_colour(d.GREEN)
```

I'm sure there is a lot more we could do here.  We should also make sure `scanner.py` and `wizard.py` keep product_id information for each of the devices to help with easier mapping.  

Is this something you would be willing to contribute to help get started?  I recommend we start small/simple, perhaps with just the dictionary and some lookup functions (a subset of my "look up" examples above).

fhempy:
Great to see that you are also interested in this functionality. I would just slightly adapt it:

> ```python
> from tinytuya import Products
> 
> product_id = 'MShdslm9Uw7Q59nN'
> 
> # Look up attributes for product
> a = Products.attributes(product_id)
> print("Device version = %d - name = %s - type = %s" % (a['version'], a['name'], a['type']))
> ```
[...]

PR #188, opened 2022-10-04
Constructor and documentation updates

uzlonewolf:
After that last PR I realized I forgot to update Device() to make the address optional, and the documentation was wrong in a couple spots.  After fixing that I decided to make persist= an optional argument for everything (it was already an option for my ThermostatDevice).  While working on that I decided I hated how a bunch of stuff was handling __init__(), so I deleted it in everything that didn't actually need it and made the rest dynamically pass everything through (while updating any defaults as needed).

Finally I condensed the constructor documentation to just OutletDevice.py and \_\_init\_\_.py and changed everything else to "See OutletDevice"

### DIFF
--- a/tinytuya/BulbDevice.py
+++ b/tinytuya/BulbDevice.py
@@ -7,11 +7,8 @@
  For more information see https://github.com/jasonacox/tinytuya
 
  Local Control Classes
-    BulbDevice(dev_id, address, local_key=None, dev_type='default')
-        dev_id (str): Device ID e.g. 01234567891234567890
-        address (str): Device Network IP Address e.g. 10.0.1.99
-        local_key (str, optional): The encryption key. Defaults to None.
-        dev_type (str): Device type for payload options (see below)
+    BulbDevice(...)
+        See OutletDevice() for constructor arguments
 
  Functions
     BulbDevice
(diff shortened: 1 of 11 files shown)


uzlonewolf:
As part of the groundwork for #185, DEVICEFILE_SAVE_VALUES now contains a list of keys which, if they exist in the Cloud.getdevices() result, are copied to devices.json.  I currently have the list set to `('category', 'product_name', 'product_id', 'biz_type', 'model', 'sub', 'icon', 'version', 'last_ip')`.  `version` and `last_ip` are currently not set anywhere but will eventually be populated by the scanner.

wizard.py seemed to have re-implemented half of tinytuya.Cloud, so that was cleaned up and now tinytuya.Cloud.getdevices() is the only code path to get the device list.

jasonacox:
Welcome back @uzlonewolf !!!  😁 

These changes make sense and the approach is far more elegant. During the review, I thought through any possible backward compatibility issue with the API and didn't see anything, though BulbDevice had me wondering.  It looks great.

Thanks for the wizard cleanup (the de-dup)!  That bugged me but never had time to clean it up.  And I love the `Cloud.getdevices()` updates for the extended device data details to `devices.json`!  I also like the clean approach with `device_info()` as the API to all devices.json details.

I feel like I need to do a good rewrite of the README to make these enhancement visible. 

Thank you! 🙏

PR #189, opened 2022-10-04
Add Climate device module for portable air conditioners. 

fr3dz10:
I tested only the Fral Supercool 19.2SC, but it seems it should support more devices. For example this one use the same DPS:
https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/eberg_qubo_q40hd_heatpump.yaml

### DIFF
--- a/tinytuya/Contrib/ClimateDevice.py
+++ b/tinytuya/Contrib/ClimateDevice.py
@@ -1,0 +1,151 @@
+from tinytuya.core import Device
+
+"""
+ Python module to interface with Tuya Portable Air Conditioner devices
+
+ Local Control Classes
+    ClimateDevice(dev_id, address, local_key=None, dev_type='default', version=3.3)
+
+ Functions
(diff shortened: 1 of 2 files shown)


jasonacox:
Thanks @fr3dz10 !  This is a good addition.

jasonacox:
@fr3dz10 Would you mind submitting a PR or posting here the [README](https://github.com/jasonacox/tinytuya/blob/master/tinytuya/Contrib/README.md) edits to describe the module?

Format something like this:

### DoorbellDevice

* DoorbellDevice - A community-contributed Python module to add support for Tuya WiFi doorbells.
* Author: [JonesMeUp](https://github.com/jonesMeUp)
* Note: Most doorbells will not stay online (to preserve battery) so controlling them locally is difficult.


    ```python
    import tinytuya
    from tinytuya.Contrib import DoorbellDevice
```

Issue #206, opened 2022-10-28
OutletThermostatDevice - Version 3.4 - how to start to create a new contrib module


Poil:
Hi,

I just bought an "ATORCH-Temperature Controller (S1TW)", it's an Outlet and a Thermostat
We can use it
* as a normal socket, switch on/off
* as a heating "socket", switch on if current temp is less than defined min temp, off if greater than defined max temp
* as a colding "socket", switch off if current temp lower than defined min temp, on if greater than defined max temp

```
    {
        "name": "chauffage bureau",
        "id": "xxxxxxxxxxx",
        "key": "xxxxxxxx",
        "mac": "fc:67:xx:xx:xx:xx",
        "category": "wk",
        "product_name": "ATORCH-Temperature Controller (S1TW)",
        "product_id": "rj84vmyhm0dh0jwy",
        "biz_type": 0,
        "model": "S1TW",
        "sub": false,
        "icon": "https://images.tuyaeu.com/smart/icon/bay1598402984069otRu/16527533308169f70b106.jpg"
    },
```

Payload, what I can extract :

```
{'dps': 
{
  '101': 'hot',     # { 'name': 'hot', 'enum': ['socket', 'hot', 'colding'] },
  '102': 206,     # { 'name': 'current_temp', 'decode': int }
  '103': False,  # { 'name': 'power_switch_status', 'decode': bool }
  '104': 0,        # { 'name': 'heating_start_temp', 'decode': int}, 
  '105': 0,         # { 'name': 'cooling_start_temp', 'decode': int}, 
  '106': 210,     # { 'name': 'heating_stop_temp', 'decode': int}, 
  '107': 0,         # { 'name': 'cooling_stop_temp', 'decode': int}, 
  '108': 0,          # { 'name': 'current_intensity', 'decode': int}, 
  '109': 0,          # { 'name': 'current_power', 'decode': int}, 
  '110': 23456,  # { 'name': 'current_voltage', 'decode': int }
  '111': 36,        # { 'name': 'total_kwh', 'decode': int }
  '112': 100,      # { 'name': 'kwh_cost', 'decode': int }
  '113': 2650,    # { 'name': 'over_voltage_limit', 'decode': int }
  '114': 2500,   # { 'name': 'over_intensity_limit', 'decode': int }
  '115': 7000,   # { 'name': 'over_power_limit', 'decode', int }
  '116': False,   # {'name': 'child_lock', 'decode': bool }
[...]

PR #213, opened 2022-11-16
Add AtorchTemperatureController

Poil:
First try #206

Not sure if this device is generic or not


jasonacox:
Thanks for the contribution @Poil !

Issue #214, opened 2022-11-18
Door/Window Sensors - "ip": 0


bernd-wechner:
I tried finding my door/window sensors with the scanner and failed. I went through the whole process of using the wizard (egads, it was time-consuming finding my devices on the Tuya site, but I got there, it was down to getting the right data centre and using manual not automatic linking before I finally found them) and I now have them all in `snapshot.json` but Wizard failed to poll them reporting `Error: No IP found` (only for these sensors, it works for a power switch I have) so I scanned the code and found the clue, and checked snapshot.json, and they appear there with `"ip": 0`. Now they do have IP addresses, and I know what they are (as I assigned them static DHCP leases).

Now, I doubt there is anything that can be done about this easily. Because I suspect the reason is simple enough, on reflection. Namely, to extend battery life, these sensors sleep waking briefly to transmit the reed switch state changes. In fact, there's a quick red LED flash shortly after every such state change, which is probably the WiFi transmission which it sleeps again.

But of course, the question remains open, and a curio indeed, if at the moment it is transmitting it exposes a UDP port (6666 or 6667) for polling or not as one does wonder if there isn't a necessary hook these have for keeping the WiFi link up for a conversation (I mean for OTA firmware updates for example or configuration changes - like binding to a new WAP etc).

That's a big **if**, of course (as in, it would take some experimenting to determine if such is in fact the case), but if it is the case, it would be interesting to support it somehow. The IP address if known could be manually added to snapshot.json, for example, and then I wonder if it's possible to watch for that IP address to appear -- which may predicate running on the gateway router or at least at a mitm site (man in the middle) as I doubt the broadcast on the LAN, but send to the Tuya data centre, and you'd need to see packets fly by from the known IP ... to then quickly check those two ports for a reply.

All rather speculative and maybe a pipe dream. I do note, there's a scanner rewrite in PR and a door sensor improvement idea already: https://github.com/jasonacox/tinytuya/issues/108.

Consider this amusing, free to close if it's not remotely tenable, or leave open if there's a prospect of maybe catching door sensor IPs and polling them if lying in wait for their transmission to fly by ...

jasonacox:
[...]

PR #219, opened 2022-11-22
Add fetching device logs to Cloud, and generic URLs

uzlonewolf:
In addition, I reworked the error handling a bit.  I usually use tinytuya.json to pass the cloud login info, but I forgot to copy it in after creating a new branch and Cloud.\_\_init\_\_() immediately blew up with `TypeError: __init__() should return None, not 'dict'` because a dict cannot be returned by \_\_init\_\_.  I wasn't sure if just setting the new .error variable would be sufficient, so I currently have it throwing a more descriptive error.  This also mimics existing behavior better.

.\_gettoken() now sets .error (and returns it) on error (such as bad auth or wrong region) instead of setting .token to an error_json object.  I considered but ultimately decided not to set .error in the various get...() functions as a bad device_id or parameter isn't necessarily fatal, unlike a .\_gettoken() error.

Included in the PR are 2 new functions, getdevicelog() and cloudrequest().  cloudrequest() just offers an easy way of fetching one of the many URLs listed in https://developer.tuya.com/en/docs/cloud/device-connection-service?id=Kb0b8geg6o761

Assuming your credentials are in tinytuya.json, you can pull the last 24 hours of logs for a device by simply:
```python
import tinytuya
import json

c = tinytuya.Cloud()
r = c.getdevicelog( '00112233445566778899' )
print( json.dumps(r, indent=2) )
```
Tuya's servers hold one week worth of logs IIRC.

Inspired by #214

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -12,12 +12,15 @@
  Functions
     Cloud
         setregion(apiRegion)
+        cloudrequest(url, action=[POST if post else GET], post={}, query={})
         getdevices(verbose=False)
         getstatus(deviceid)
         getfunctions(deviceid)
         getproperties(deviceid)
         getdps(deviceid)
         sendcommand(deviceid, commands)
+        getconnectstatus(deviceid)
+        getdevicelog(deviceid, start=[now - 1 day], end=[now], evtype="1,2,3,4,5,6,7,8,9,10", size=100, params={})
 """
 
 from .core import *
@@ -73,9 +76,10 @@ class Cloud(object):
         self.urlhost = ''
         self.uid = None     # Tuya Cloud User ID
         self.token = None
+        self.error = None
         self.new_sign_algorithm = new_sign_algorithm
 
-        if apiKey is None or apiSecret is None:
+        if (not apiKey) or (not apiSecret):
             try:
                 # Load defaults from config file if available
                 config = {}
@@ -86,14 +90,16 @@ class Cloud(object):
                     self.apiSecret = config['apiSecret']
                     self.apiDeviceID = config['apiDeviceID']
             except:
-                return error_json(
-                    ERR_CLOUDKEY,
-                    "Tuya Cloud Key and Secret required",
-                )
+                #self.error = error_json(
+                #    ERR_CLOUDKEY,
+                #    "Tuya Cloud Key and Secret required",
+                #)
+                #return
+                raise ValueError('Tuya Cloud Key and Secret required')
 
         self.setregion(apiRegion)
         # Attempt to connect to cloud and get token
-        self.token = self._gettoken()
+        self._gettoken()
 
     def setregion(self, apiRegion=None):
         # Set hostname based on apiRegion
@@ -112,19 +118,49 @@ class Cloud(object):
         if self.apiRegion == "in":
             self.urlhost = "openapi.tuyain.com"      # India Datacenter
 
-    def _tuyaplatform(self, uri, action='GET', post=None, ver='v1.0', recursive=False):
+    def _tuyaplatform(self, uri, action='GET', post=None, ver='v1.0', recursive=False, query=None):
         """
         Handle GET and POST requests to Tuya Cloud
         """
         # Build URL and Header
-        url = "https://%s/%s/%s" % (self.urlhost, ver, uri)
+        if ver:
+            url = "https://%s/%s/%s" % (self.urlhost, ver, uri)
+        elif uri[0] == '/':
+            url = "https://%s%s" % (self.urlhost, uri)
+        else:
+            url = "https://%s/%s" % (self.urlhost, uri)
         headers = {}
         body = {}
-        if action == 'POST':
+        sign_url = url
+        if post is not None:
             body = json.dumps(post)
             headers['Content-type'] = 'application/json'
-        else:
-            action = 'GET'
+        if action not in ('GET', 'POST', 'PUT', 'DELETE'):
+            action = 'POST' if post else 'GET'
+        if query:
+            # note: signature must be calculated before URL-encoding!
+            if type(query) == str:
+                # if it's a string then assume no url-encoding is needed
+                if query[0] == '?':
+                    url += query
+                else:
+                    url += '?' + query
+                sign_url = url
+            else:
+                # dicts are unsorted, however Tuya requires the keys to be in alphabetical order for signing
+                #  as per https://developer.tuya.com/en/docs/iot/singnature?id=Ka43a5mtx1gsc
+                if type(query) == dict:
+                    sorted_query = []
+                    for k in sorted(query.keys()):
+                        sorted_query.append( (k, query[k]) )
+                    query = sorted_query
+                    # calculate signature without url-encoding
+                    sign_url += '?' + '&'.join( [str(x[0]) + '=' + str(x[1]) for x in query] )
+                    req = requests.Request(action, url, params=query).prepare()
+                    url = req.url
+                else:
+                    req = requests.Request(action, url, params=query).prepare()
+                    sign_url = url = req.url
         now = int(time.time()*1000)
         headers = dict(list(headers.items()) + [('Signature-Headers', ":".join(headers.keys()))]) if headers else {}
         if self.token is None:
@@ -140,7 +176,7 @@ class Cloud(object):
                 ''.join(['%s:%s\n'%(key, headers[key])                                   # Headers
                             for key in headers.get("Signature-Headers", "").split(":")
                             if key in headers]) + '\n' +
-                '/' + url.split('//', 1)[-1].split('/', 1)[-1])
+                '/' + sign_url.split('//', 1)[-1].split('/', 1)[-1])
         # Sign Payload
         signature = hmac.new(
             self.apiSecret.encode('utf-8'),
@@ -175,8 +211,8 @@ class Cloud(object):
                 log.debug("Failed 2nd attempt to renew token - Aborting")
                 return None
             log.debug("Token Expired - Try to renew")
-            token = self._gettoken()
-            if "err" in token:
+            self._gettoken()
+            if not self.token:
                 log.debug("Failed to renew token")
                 return None
             else:
@@ -188,10 +224,11 @@ class Cloud(object):
             try:
                 response_dict = json.loads(response.content)
             except:
-                return error_json(
+                self.error = error_json(
                     ERR_CLOUDKEY,
                     "Cloud _tuyaplatform() invalid response: %r" % response.content,
                 )
+                return self.error
         # Check to see if token is expired
         return response_dict
 
@@ -200,17 +237,20 @@ class Cloud(object):
         self.token = None
         response_dict = self._tuyaplatform('token?grant_type=1')
 
-        if not response_dict['success']:
-            return error_json(
+        if not response_dict or 'success' not in response_dict or not response_dict['success']:
+            self.error = error_json(
                 ERR_CLOUDTOKEN,
                 "Cloud _gettoken() failed: %r" % response_dict['msg'],
             )
+            return self.error
 
         self.token = response_dict['result']['access_token']
         return self.token
 
     def _getuid(self, deviceid=None):
         # Get user ID (UID) for deviceid
+        if self.error:
+            return self.error
         if deviceid is None:
             return error_json(
                 ERR_PARAMS,
@@ -235,6 +275,20 @@ class Cloud(object):
         uid = response_dict['result']['uid']
         return uid
 
+    def cloudrequest(self, url, action=None, post=None, query=None):
+        """
+        Make a generic cloud request and return the results.
+
+        Args:
+          url:    Required.  The URL to fetch.
+          action: Optional.  GET, POST, DELETE, or PUT.  Defaults to GET, unless POST data is supplied.
+          post:   Optional.  POST body data.  Will be fed into json.dumps() before posting.
+          query:  Optional.  A dict containing query string key/value pairs.
+        """
+        if action is None:
+            action = 'POST' if post else 'GET'
+        return self._tuyaplatform(url, action=action, post=post, ver=None, query=query)
+
     def getdevices(self, verbose=False):
         """
         Return dictionary of all devices.
@@ -245,7 +299,7 @@ class Cloud(object):
         if uid is None:
             return error_json(
                 ERR_CLOUD,
-                "Unable to get device list"
+                "Unable to get uid for device list"
             )
         elif isinstance( uid, dict):
             return uid
@@ -386,3 +440,46 @@ class Cloud(object):
                     "Error from Tuya Cloud: %r" % response_dict['msg'],
             )
         return(response_dict["result"]["online"])
+
+    def getdevicelog(self, devid, start=None, end=None, evtype=None, size=100, params={}):
+        """
+        Get the logs for a device.
+
+        Note: The cloud only returns logs for DPs in the "official" DPS list.
+          If the device specifications are wrong then not all logs will be returned!
+          This is a limitation of Tuya's servers and there's nothing we can do about it.
+
+        Args:
+          devid:  Required.  Device ID
+          start:  Optional.  Get logs starting from this time.  Defaults to yesterday
+          end:    Optional.  Get logs until this time.  Defaults to the current time
+          evtype: Optional.  Limit to events of this type.  1 = Online, 7 = DP Reports.  Defaults to all events.
+          size:   Optional.  Maximum number of log entries to return.  Defaults to 100.
+          params: Optional.  Additional values to include in the query string.  Defaults to an empty dict.
+
+        Returns:
+          Response from server
+        """
+        if not end:
+            end = int(time.mktime(time.gmtime()))
+        if not start:
+            start = end - 86400
+        if not evtype:
+            # get them all by default
+            # 1 = device online, 7 = DP report
+            # https://developer.tuya.com/en/docs/cloud/0a30fc557f?id=Ka7kjybdo0jse
+            evtype = '1,2,3,4,5,6,7,8,9,10'
+        if type(params) != dict:
+            params = {}
+        if 'start_time' not in params:
+            params['start_time'] = start
+        if 'end_time' not in params:
+            params['end_time'] = end
+        if 'type' not in params:
+            params['type'] = evtype
+        if 'size' not in params:
+            params['size'] = size
+        if 'query_type' not in params:
+            params['query_type'] = 1
+
+        return self.cloudrequest( '/v1.0/devices/%s/logs' % devid, query=params)
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -90,12 +90,12 @@ class Cloud(object):
                     self.apiSecret = config['apiSecret']
                     self.apiDeviceID = config['apiDeviceID']
             except:
-                #self.error = error_json(
-                #    ERR_CLOUDKEY,
-                #    "Tuya Cloud Key and Secret required",
-                #)
-                #return
-                raise ValueError('Tuya Cloud Key and Secret required')
+                self.error = error_json(
+                    ERR_CLOUDKEY,
+                    "Tuya Cloud Key and Secret required",
+                )
+                #return self.error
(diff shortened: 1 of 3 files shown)


jasonacox:
Brilliant! Thanks @uzlonewolf. The improved error handling and cleanup is also much appreciated. 🙏 

I'll run a few tests and release this as a minor update. I think this Cloud API enhancement is worth getting out quickly.  It may be good to include your code example in the Cloud section of the main README.

jasonacox:
PR changes now live in [v1.8.0](https://pypi.org/project/tinytuya/1.8.0/).

bernd-wechner:
Fantastic, thanks for this. Works a charm. Now just have to work out how to interpret those logs.

bernd-wechner:
Best I've found is this:

https://developer.tuya.com/en/docs/cloud/cbea13f274?id=Kalmcohrembze

Alas it's a bit spartan. For example I get a lot of event ID 7 and the doc:



event_id | Integer | Type of event: 1  online, 2 offline, 3 device activation, 4 device reset, 6 firmware  upgrade, 8 device semaphore, 9 device restart, 10 timing information
-- | -- | --


The event time is clearly encoded, but not documented how:



event_time | Long | Event Time
-- | -- | --

And I'm seeing events like:

{'event_from': '1', 'event_id': 1, 'event_time': 1669846757908, 'status': '1'}

and:

{'code': 'doorcontact_state', 'event_from': '1', 'event_id': 7, 'event_time': 1669846737501, 'status': '1', 'value': 'true'}

and trying to correlate them with what I see in their web interface. I'll plod along but just though to drop a line in case any gems of wisdom fall from the cloud to help me ;-)

uzlonewolf:
Yeah, the documentation leaves a lot to be desired.  Higher up in the page it says 7=data report.  The times are UNIX timestamp * 1000, so 1669846737501 = 1669846737.501 = 2022-11-30 14:18:57

bernd-wechner:
This is great and am loving it! Making huge headway into ensuring all the data is in a local database ;-). But as I retire for the night, I'm stuck on one problem the docs don't seem to say much on. Suddenly, for the first time, I see the result contains:

`'has_next': True,`

and

`'next_row_key': '<long key>`

What to do with this? 

Could/should `getdevicelog` handle this, simply fetching the next row until has_next is False?

uzlonewolf:
> Could/should `getdevicelog` handle this, simply fetching the next row until has_next is False?

That is a very good question.  I think it depends on whether or not `getdevicelog` is returning all `size=` records; if it is, then it's the caller's problem, otherwise `getdevicelog` should probably do it.

I should probably add an argument to `getdevicelog` for that row key, but in the meantime `getdevicelog( <same args as original call>, params={"last_row_key": "<long key from next_row_key>"})` should work.

How often are you pulling logs? An issue I didn't even think about when writing this is #230.

bernd-wechner:
Ouch.  Not clear to me how that limit works. Described in 230 as a flat limit and a rate limit? 

The latter is viable.  The former kills my plans and pushes me back to considering reflash options.  

I want a lasting solution.  

I've done a few fetches to test of course (10s thereof) and long term need only one fetch per day ideally (as long as it returns 24 hours of data)

bernd-wechner:
Only 4 door sensors. So it sounds like the limits will be fine, either way.  Will check the multi row thing when next able (at the workstation).

bernd-wechner:
> I think it depends on whether or not `getdevicelog` is returning all `size=` records; if it is, then it's the caller's problem, otherwise `getdevicelog` should probably do it.

Looked into this quickly and doesn't look like TinyTuya has anywhere the logic to collect all the rows of a log. To evidence that:

```
bernd@bigfoot:~/workspace/IoT$ git clone https://github.com/jasonacox/tinytuya.git
Cloning into 'tinytuya'...
remote: Enumerating objects: 2181, done.
remote: Counting objects: 100% (11/11), done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 2181 (delta 4), reused 10 (delta 4), pack-reused 2170
Receiving objects: 100% (2181/2181), 1.04 MiB | 2.82 MiB/s, done.
Resolving deltas: 100% (1350/1350), done.
bernd@bigfoot:~/workspace/IoT$ grep -R has_next tinytuya
bernd@bigfoot:~/workspace/IoT$ grep -R next_row_key tinytuya
```
I shall try with adding `params={"last_row_key": "<long key from next_row_key>"}` to the args and fetch them one by one, and maybe write me a wrapper that does that (which just gets me the whole log). There have to be enough for has_next is true, which I've only seen once so far. But if I catch it, I'll spring to action and test this idea out.

uzlonewolf:
> There have to be enough for has_next is true, which I've only seen once so far. But if I catch it, I'll spring to action and test this idea out.

Yes, it should only happen if there are more than `size=` records for the given time range (default is `size=100`).  You can force it by specifying a small number such as `size=5`.

uzlonewolf:
Ok, I worked on this a bit this morning and discovered a few things.  It turns out https://developer.tuya.com/en/docs/cloud/0a30fc557f?id=Ka7kjybdo0jse is the documentation for the URL we're using; the one linked earlier in this thread is for a slightly different URL that we do not have access to (`/v1.0/devices/` vs `/v1.0/iot-03/devices/`; the latter returns `"msg": "No permissions. This API is not subscribed."` when called).  They're similar but the query parameters and returned values are slightly different.  As for the returned logs, there is a slight issue I forgot about: if the "official" DPS map is wrong, the DP's not in the official map are counted but filtered out by the server and not returned.  This means it can set `'has_next': True` while returning fewer results than requested.  Finally, the param to get the next set is `start_row_key` and should be set to what was returned in `next_row_key`.  I'll put together a PR that handles pulling the rest when 'has_next' is True and fewer results than requested are returned.

bernd-wechner:
Thanks heaps.  It's not a big code change and I'd happily have done it too, but it's hard to do unless you can see has_next as true to test (an untested code snippet,  even small is pretty risky).  And I've only seen it once to date.

uzlonewolf:
Finally finished writing the loop.  With it retrying to get everything, my biggest concern is it somehow getting stuck in a loop and chewing through all your monthly requests, so I have limited it to at most 50 tries.  Combined with the server's hard limit of 100 results per try, this means you can pull at most 50*100=5000 log messages per function call.  For some extremely chatty devices, such as the thermostats I have, even that may not be enough to pull them all, so you might need to either also loop yourself or set the new `max_fetches=` parameter to something greater than 50.  The `size=` parameter is now an approximate target and the actual number of logs returned will be somewhere between 0 and `size * 2 - 1`; set `size=0` to get up to `max_fetches * 100` entries.  I also added a `start_row_key=` parameter while I was at it.  The new function definition is
`getdevicelog(deviceid=None, start=None, end=None, evtype=None, size=0, start_row_key=None, max_fetches=50, params={})`

bernd-wechner:
Interesting. Some thoughts:

1. An endless loop would only be a risk surely if the next_row_key pointed back to an already fetched log (i.e. a bug at their server end). But that could be guarded against by keeping a set of already fetched row_keys and detecting such a loop.

2. `max_fetches` is a sensible option yes, though I'd differentiate between fetching next rows and retrying a fetch that failed, or? I typically code up a `max_retries` parameter for retries (and 50 is generous of that, I usually settle on 5 to 10, but it depends on the reliability of the API I'm using).

3.  The date range should of course be honoured server side, and limit the number of rows, but can also, if keen, be checked client side. I observed for example that the logs seem to be bundled reverse chronology (most recent event first, the oldest event last), and so client side easy to check that the first event is at or before the requested end time and also stop fetching if a log ever contains a last event at or prior to the start time. 

4. For now, I am fetching the whole log and actually do that by setting the `start=1, end=sys.maxsize` as I'm testing manually, but I'll set up in time, a daily fetch in time for the last day each time. I might even sensibly write min to start at the time of the last fetched event to now on each run and assert that the last event in the returned log is one I've already seen. As I intend to record all these events in a database locally.

uzlonewolf:
Oh yeah, as for 3.  I'm again going to point to my thermostats.  Unfortunately the official DPS map for them isn't even close and 70% of the DP's they use are not listed (and are thus filtered out from the server response).  In my testing I've seen gaps of 300+ log entries (3+ fetches) where the returned logs array is completely empty and the only clue it's not stuck is the next_row_key changing.  As such, timestamp checking on the client side won't be reliable for devices like these.

bernd-wechner:
Remind me what's a DP?

uzlonewolf:
Data Points (DP/DPS).  It's the numeric ID # Tuya devices use to identify each input or output.  You need to know the ID # when talking to devices locally, however the cloud translates them into names.  I.e. for my thermostats the current temperature in degrees F is DPS 29 however the cloud reports it as `temp_current_f`.

bernd-wechner:
Is that the same as the `deviceid` argument to `getdevicelog()`? And the `device_id` element of the returned result structure?

uzlonewolf:
No, the device id identifies the device, while the DPS identifies the individual inputs/outputs on the device.  I.e. a 4-channel relay controller will have a device_id for the device itself, plus 4 DP's for the 4 different outputs (usually DPS 1-4).  Most devices will have many more DPs than that for things like countdown timers, temperature sensors, current sensors, etc.

bernd-wechner:
Hmmm, but confused. Is the DP in the returned response a proper of each event? If so, where is it here:

https://github.com/jasonacox/tinytuya/issues/214#issuecomment-1321043692

Forgive me, just trying to get my head around all this.

uzlonewolf:
Yes and no.  The cloud maps the DPS IDs to names they call "code."

Using that linked comment as an example, if I wanted to change the "muffling" value using a local connection I would need to use the numeric DPS ID:

```python
d = tinytuya.Device( '...', '...' )
d.set_value(16, False) # Turn off DPS 16 ("muffling")
```

To get the DPS<->code mapping you can use `c.getdps( device_id )`

```json
{
  "result": {
    "category": "ywbj",
    "functions": [
      {
        "code": "muffling",
        "dp_id": 16,
        "type": "Boolean",
        "values": "{}"
      }
    ],
    "status": [
      {
        "code": "smoke_sensor_status",
        "dp_id": 1,
        "type": "Enum",
        "values": "{\"range\":[\"alarm\",\"normal\"]}"
      },
      {
        "code": "smoke_sensor_value",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"\",\"min\":0,\"max\":100,\"scale\":1,\"step\":1}"
      },
      {
        "code": "battery_state",
        "dp_id": 14,
        "type": "Enum",
        "values": "{\"range\":[\"low\",\"middle\",\"high\"]}"
      },
      {
        "code": "battery_percentage",
        "dp_id": 15,
        "type": "Integer",
        "values": "{\"unit\":\"%\",\"min\":0,\"max\":100,\"scale\":0,\"step\":1}"
      },
      {
        "code": "muffling",
        "dp_id": 16,
        "type": "Boolean",
        "values": "{}"
      }
    ]
  },
...
}
```

But this is the gotcha with pulling cloud logs: half of the devices I have do not have a proper DPS<->code mapping like that.  Those DP's will not show up in the event logs because the server cannot map the DPS ID to a code.

bernd-wechner:
Ah, thanks for clarifying. "code" is such a weird element IMHO. Sometimes I see it as empty, too. So what I'm hearing is they have a number to string mapping for code, as they do for event IDs and Sources, which I have mapped as:

```
EVENT_IDS = {1: "online",
             2: "offline",
             3: "device activation",
             4: "device reset",
             5: "command issuance",
             6: "firmware upgrade",
             7: "data report",
             8: "device semaphore",
             9: "device restart",
             10: "timing information"}

EVENT_SOURCES = {-1: "unknown",
                  1: "device itself",
                  2: "client instructions",
                  3: "third-party platforms",
                  4: "cloud instructions"}
```

But unlike Event IDs and sources the logs don't return the numeric, they return the mapped string for `code` and sometimes they have no mapping implemented?

uzlonewolf:
Yes, exactly.

bernd-wechner:
> One page says you get "about 26,000 API calls" but another says Cloud API calls are "Free of charge within monthly usage limits of $ 0.20". I'm leaning towards the limit is per month.

Trying to find the docs on this. Did you keep links? I'm curious to see where they state this and what one fetch costs (toward a possible $0.20 monthly gratis limit).

uzlonewolf:
Yeah, they were posted in #230.  https://developer.tuya.com/en/docs/iot/membership-service?id=K9m8k45jwvg9j gives the 26k figure and https://www.tuya.com/vas/commodity/IOT_CORE_V2 -> "Trial Addition" -> "View More" gives $0.20.

bernd-wechner:
Thanks! The first page is mystical indeed, and on the second I see:

image

No "Trial Addition" but I see "Trial Edition (one month)" which is selected but I don't see "View More" anywhere.

You're right though Tuya are clear as mud.

uzlonewolf:
If you mouse-hover over "Flagship" or "Corporate" you get a pop-up with a "View More" link.  I could have swore that same pop-up existed for the "Trial" option as well.  I wonder if they removed it for some reason.

uzlonewolf:
I was able to find a reference to $0.20 by going to iot.tuya.com -> Cloud -> Cloud Services and clicking "View Details" for "IoT Core."  Under "Usage/Resource Pack Quota" it says "0 / 0.2 USD" while "Quota Refresh" says "Monthly."

bernd-wechner:
Thanks enormously, that is so cool.  Have boiled it down to:

https://developer.tuya.com/en/docs/iot/membership-service?id=K9m8k45jwvg9j

where is written:

    Trial Edition (1 month)
    Note: Only for individual developers or use in the debugging phase, and commercial use is prohibited
        50 devices
        Trial resource pack: about 26 thousand API calls
        When developers exceed the limit, Tuya will stop the service.

and later on same page:

    Trial
    3.71 USD/million API calls (for Plan 1 which applies us in Australia)
    About 26 thousand API calls
 
And then on 

https://eu.iot.tuya.com/cloud/products/detail?abilityId=1442730014117204014&id=p1668767995023hmaagk&abilityAuth=0&tab=1

it is written:

    Resource Pack Name                   Usage/Resource Pack Quota    Quota Refresh    Effective Date         Expiration Date        Status
    Cloud Develop Base Resource Trial    0.002176 / 0.2 USD           Monthly          2022-11-18 21:42:25    2022-12-18 21:42:25    In service

And 0.2 USD at 3.71USD/million API calls translates with a little math to 53,908 API calls per month or 1739 API calls/day.

Which is indeed I agree, the most reliable snippet as it is:

1. Only visible when logged in, so specific to my (your) account
2. Labelled "My subscriptions"
3. Clearly states the quota is refreshed Monthly 

Very encouraging and all noted in the internal doc of the app I'm putting together now.

bernd-wechner:
Ouch, I just got: "No permissions. Your subscription to cloud development plan has expired." Now that was a surpise.

bernd-wechner:
And I can't see any reason on their website. None at all. I can surf to this:

image

and if I try to buy the free 1 month trial I can't, this pops up: 

image

And of course the Flagship and Corporate edition are ridiculously expensive (like the price of a car). Am I now without services, after all that research we put into the rate limits and the Base Resource Trial above being a Monthly Quota Refresh.

uzlonewolf:
Ooof, hopefully it's temporary and resets at the end of the month.  I'm many months past the "one month" trial and they haven't shut me off yet.

tinytuya/Cloud.py
```diff
@@ -386,3 +451,46 @@ def getconnectstatus(self, deviceid=None):
                     "Error from Tuya Cloud: %r" % response_dict['msg'],
             )
         return(response_dict["result"]["online"])
+
+    def getdevicelog(self, devid, start=None, end=None, evtype=None, size=100, params={}):
```
klightspeed, line 455: This default value for `params` is a mutable static variable (refer <https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments>), and as a result carries over its values from the previous request within the same python program, resulting in responses with zero entries when requesting a second device's log after previously retrieving the logs for another device.

jasonacox: Thanks @klightspeed !  Pylint also caught this and I forgot to ask @uzlonewolf .  

@uzlonewolf what if set this default to None and add a...
```python
if not params:
     params = {}
```

klightspeed: It looks like the
```py
        if type(params) != dict:
            params = {}
```
should already cover if the default value for `params` were `None`.
```py
>>> type(None) != dict
True
```

uzlonewolf: Which leads to the question, since the dict is modified, should we be doing `params = params.copy()` as well?  Keeping the start/end times the same between requests could be useful (or confusing), but changing the size or start_row_key is not.  We could also save/restore just those keys.

I'm leaning towards leaving the default as it is and instead going
```python
if type(params) != dict:
    params = {}
else:
    params = params.copy()
```

jasonacox: This hurt my head a bit. I had to run a few test to prove this out and educate myself. :)

In the current state, when getdevicelog() is called.. 

* without params provide, it will set params with the keys (e.g. start_time) that will leak into the global/calling scope so that when it is called again w/o params, those previous values will still be there (the static aspect you talk about).  I'm not sure why we would want this (agree with "confusing").  Can you think of a reason why we would want that?  I can't but I may be missing something.  In any case, both recommendations above will remove this scope leaking. 
* with params specified, it will only modify the params sent and they will not leak to others.  I did some testing, and what I see is that if you provide explicit (static) params (eg. {'id': 123} ), the additional keys added are not leaked back to the calling scope.  However, if you call with a variable that contains the params dict, the additional updates are preserved (in the variable) and will be seen in the calling scope.

So both the recommendation from @klightspeed and @uzlonewolf result in preserving the second point above and fixing the first point.  If no params are specified in the calling scope, nothing is leaked back.

@uzlonewolf I know you are leaning toward leaving the `params={}` default, but if we just change it to `params=None`, we don't even need to add your conditional setting.  The existing conditionals will handle setting params to a empty dict if no params were specified.

My testing:

```python
def f(params={}):
    print("called with %r" % params)
    params['id'] = 1234

def g(params=None):
    if type(params) != dict:
        params = {}
    print("called with %r" % params)
    params['id'] = 1234
```

```python
# EXISTING

# try with default - see the leak
>>> f()
called with {}
>>> f()
called with {'id': 1234}

# try with static params - no leak until empty params is called
>>> f({'z': 99})
called with {'z': 99}
>>> f({'z': 99})
called with {'z': 99}
>>> f()
called with {'id': 1234}

# params as a variable - expected leak until empty params is called
>>> p = {'z': 99}
>>> f(p)
called with {'z': 99}
>>> f(p)
called with {'z': 99, 'id': 1234}
>>> f()
called with {'id': 1234}

# PROPOSED

# default - no leak
>>> g()
called with {}
>>> g()
called with {}

# with static - no leak
>>> g({'z': 99})
called with {'z': 99}
>>> g({'z': 99})
called with {'z': 99}
>>> g()
called with {}

# params as a variable - expected leak but no leak when empty params is called
>>> p = {'z': 99}
>>> g(p)
called with {'z': 99}
>>> g(p)
called with {'z': 99, 'id': 1234}
>>> g()
called with {}
```

uzlonewolf: > we don't even need to add your conditional setting

We do if we don't want `start_row_key` to leak back in the case of `params` being provided and multiple calls are needed to get all requested records, or `size` to get added unexpectedly.

```python
p = {}
c.getdevicelog( ..., start_row_key="...", parms=p) # if either start_row_key is provided or this needs multiple requests to fetch all rows, then
print(p) # this now contains p['start_row_key'] and
c.getdevicelog( ..., parms=p) # will unexpectedly try and start with the above start_row_key
```
Same deal with `size`
```python
p = {}
c.getdevicelog( ..., size=1, parms=p)
print(p) # this now contains p['size'] = 1 and
c.getdevicelog( ..., size=100, parms=p) # will unexpectedly retrieve at most 1 record per call, resulting in 50 cloud calls and (at most) 50 records returned
```

jasonacox: Ha!  You're right!  I thought about it just as I was going to sleep last night that we actually may not want this to leak back into the passed variable. You just proved it.

I'll submit a change.  Thanks for spotting this @klightspeed and thanks for finding/fixing he real issue @uzlonewolf .  Brilliant work you two.

PR #220, opened 2022-11-22
Update README for new Cloud functions, and tighter deviceid error checking

uzlonewolf:
README update as requested in #219, and a bit tighter error checking for the deviceid arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,12 +208,15 @@ Functions:
     
     Cloud
         setregion(apiRegion)
+	cloudrequest(url, action=[POST if post else GET], post={}, query={})
         getdevices(verbose=False)
         getstatus(deviceid)
         getfunctions(deviceid)
         getproperties(deviceid)
         getdps(deviceid)
         sendcommand(deviceid, commands)
+	getconnectstatus(deviceid)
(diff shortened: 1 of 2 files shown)


jasonacox:
Thank you!

Off topic a bit, but you got me thinking about expanding Cloud functions, should we expose short-cut lookup functions as well?  I'm thinking about `getname(id)` `getkey(id)` `getmac(id)` etc. (though things like product_name and icon could be helpful or useless).  All of this can be derived from `getdevices()` payload today, but wondering if these functions could be useful.

uzlonewolf:
I'm not feeling functions for stuff that's available from `getdevices()`, you just know someone is going to do something stupid and make a new request for every attribute for every device, and I'd rather not give Tuya a reason to restrict API access.  If it's for something not available from `getdevices()` then sure, but with the sheer number of possible API calls I would only add them for things which would be rather useful (such as the device logs) or if we get a request for it.

uzlonewolf:
Oh, and Happy Thanksgiving!

jasonacox:
Thanks @uzlonewolf !!  Happy Thanksgiving to you too!  🙏

PR #229, opened 2022-12-02
Rework Cloud log start/end times, and update documentation

uzlonewolf:
I noticed the Cloud documentation calls for 13-digit timestamps, however I was sending 10-digit UNIX timestamps.  It seemed to work fine, but I decided it would be better to match the official documentation.  So it now sends a 13-digit `int(time.time() * 1000)`.  I also reworked it so you can pass a negative number for that many days ago (i.e. -2 means 2 days ago) and added automatic server time offset compensation.

Documentation for the above as well as the Zigbee Gateway example were added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,13 +392,15 @@ result = c.sendcommand(id,commands)
 print("Results\n:", result)
 ```
 
-Up to one week of device logs can also be pulled from the Cloud.  By default getdevicelog() will pull 1 day of logs or 100 log entries, whichever comes first.
+Up to one week of device logs can also be pulled from the Cloud.  By default getdevicelog() will pull 1 day of logs or 100 log entries, whichever comes first.  The returned timestamps are unixtime*1000, and event_id 7 (data report) will probably be the most useful.
 
 ```python
 import tinytuya
 import json
 
 c = tinytuya.Cloud()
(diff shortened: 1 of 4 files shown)


jasonacox:
Thanks @uzlonewolf !  I'm going to work on a PyPI release.

jasonacox:
I released this as [v1.9.0](https://pypi.org/project/tinytuya/1.9.0/) - Thanks @uzlonewolf!

One minor issue I see... During my testing, I'm seeing a random "0" show up during `wizard`:

```
TinyTuya Setup Wizard [1.9.0]

    Existing settings:
        API Key=abcdefghijklmnopqrstuv 
        Secret=1234567890abcdef1234567890abcdef1
        DeviceID=123abc4567890abcdef12
        Region=us

    Use existing credentials (Y/n): 
0


Device Listing
```

The culprit is the Cloud init call due to `_gettoken()`:

```python
log.warning(self.server_time_offset)
```

Since it is a warning, it is showing up under normal conditions so I'll drop it to `debug`:

```python
log.debug("server_time_offset: %r", self.server_time_offset)
```

uzlonewolf:
Ah, oops, I had intended to delete that line but it would seem I forgot.  Thanks for the catch and fix.

PR #242, opened 2022-12-20
Added delayedoff & help function to server

cowboy3d:
I was missing the functionality to turn off a device with a delay - using this enables e.g. to shutdown your pi and completely turn it off shutting off its power plug using delayedoff.

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -23,6 +23,8 @@ Description
                                         - Sets the Tuya Cloud API login info
         /offline                        - List of registered devices that are offline
 
+        /delayoff/{DeviceID}/{Time} - Turn off device with delay
+
 """
 
 # Modules
@@ -293,6 +295,11 @@ class ThreadingHTTPServer(ThreadingMixIn, HTTPServer):
     daemon_threads = True
     pass
 
+def delayoff(d, sw, delay):
+    time.sleep(int(delay))
+    d.turn_off(switch=sw, nowait=True)
+    d.close()
+
 class handler(BaseHTTPRequestHandler):
     def log_message(self, format, *args):
         if DEBUGMODE:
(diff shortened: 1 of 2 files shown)


server/server.py
```diff
@@ -386,6 +408,29 @@ def do_GET(self):
...
+                    (id, sw, delay) = id.split("/")
+                except:
+                    id = ""
+                    message = json.dumps({"Error": "Invalid syntax in delayoff command.", "url": self.path})
+            if id in deviceslist:
+                try:
+                    d = tinytuya.OutletDevice(id, deviceslist[id]["ip"], deviceslist[id]["key"])
+                    d.set_version(float(deviceslist[id]["version"]))
+
+                    thread = threading.Thread(target = delayoff, args = (d, sw, delay))
```
uzlonewolf, line 426: Any particular reason you went with threading.Thread()+sleep() instead of threading.Timer()?

server/server.py
```diff
@@ -386,6 +408,29 @@ def do_GET(self):
...
+                    thread.start()
+
```
uzlonewolf, line 428: Any particular reason you went with threading.Thread()+sleep() instead of threading.Timer()?

jasonacox: Good question @uzlonewolf . 

Thanks for the contribution @cowboy3d !  I'll accept this for now but would love to hear your thoughts on Thread() vs Timer() approach.

server/server.py
```diff
@@ -327,6 +334,21 @@ def do_GET(self):
         contenttype = 'application/json'
         if self.path == '/devices':
             message = json.dumps(deviceslist)
+        elif self.path == '/help':
```
jasonacox, line 337: This is a good addition.

jasonacox: I debated if it would be better as an HTML payload, but the JSON response still provides the help.

server/server.py
cowboy3d: Thanks a lot for accepting the merge request. I'll rework the part using threading.Timer() - should be a rather simply change.

PR #243, opened 2022-12-22
Delay off - replace thread & sleep with threading.timer

cowboy3d:
I just did a small optimization to the merged code - replaced thread & sleep with a threading.timer.


jasonacox:
Thanks @cowboy3d !

PR #244, opened 2022-12-27
Modify device dps using web browser

cowboy3d:
Hi,
I just did a small addition to the served web pages - if you click on the device id you will be routed to a new page that allows the setting of the device's dps. 
If there only would be a way to query the characteristics of the individual dps the page could be greatly improved.

### DIFF
--- a/server/web/device_dps.html
+++ b/server/web/device_dps.html
@@ -1,0 +1,143 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<title>TinyTuya API Server - Device Details</title>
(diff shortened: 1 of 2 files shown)


jasonacox:
Thanks @cowboy3d ! This is a great addition. I'm already using it to test. :) 

I agree on the DPS mapping. I would like to figure out a way to create a mapping that is actually real vs the mess you get from the Tuya Cloud definitions.  I don't want to bloat the module with a massive DB of product id and dps maps, but it could be a separate repository (or folder under this one). Key would be to figure out how to easily but accurately accept community product updates.

PR #253, opened 2023-01-13
Allow nowait for status(), and split message parsing into separate function

uzlonewolf:
I have a single-threaded application where I handle the socket reading (via select.select() and buffering the data) and needed status() to not block.  Once I get a message I then need to parse it, so I moved the parsing part of _send_receive() into its own function to make it easier to call.


jasonacox:
Logical. I can't think of any dependency issue for this change so I will rev the version as a minor patch, v1.9.2.

PR #262, opened 2023-01-17
Normalize snapshot.json fields and add "version" to devices.json

uzlonewolf:
Snapshot saving should now be done through tinytuya.scanner.save_snapshotfile().  This function normalizes all the field names and value types before saving.  To load, call tinytuya.scanner.load_snapshotfile() to convert back.

I went with "ver" and "id" in snapshot.json as that's what older versions used.  When loaded, "ver" gets replaced with "version" and "id" is copied into "gwId"

### DIFF
--- a/tinytuya/scanner.py
+++ b/tinytuya/scanner.py
@@ -314,7 +314,7 @@ class ForceScannedDevice(DeviceDetect):
             self.brute_force_v3x_data()
 
         if not self.ver_found:
-            self.deviceinfo['version'] = self.deviceinfo['ver'] = 0.0
+            self.deviceinfo['version'] = 0.0
 
         if self.options['verbose'] and self.found and not self.displayed:
             _print_device_info( self.deviceinfo, 'Failed to Force-Scan, FORCED STOP', self.options['termcolors'], self.message )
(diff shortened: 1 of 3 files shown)


tinytuya/scanner.py
```diff
@@ -1536,12 +1542,14 @@ def tuyaLookup(deviceid):
     for ip in broadcasted_devices:
         dev = broadcasted_devices[ip].deviceinfo
         dev['ip'] = ip
+        dev['origin'] = 'broadcast'
```
jasonacox, line 1545: Love this!

tinytuya/scanner.py
```diff
@@ -1615,6 +1616,73 @@ def _display_status( item, dps, term ):
...
+    if 'ver' in itm and itm['ver']:
+        itm['version'] = float(itm['ver'])
+        del itm['ver']
+    elif 'version' in itm and itm['version']:
+        itm['version'] = float(itm['version'])
+    else:
+        itm['version'] = 0.0
+    return itm
+
+def _snapshot_save_item( old ):
```
jasonacox, line 1631: Perfect

tinytuya/scanner.py
```diff
@@ -1615,6 +1616,73 @@ def _display_status( item, dps, term ):
             print("    %s[%-25.25s] %s%-18s - DPS: %r" %
                   (term.subbold, name, term.dim, ip, dps))
 
+def _snapshot_load_item( itm ):
+    # normalize all the fields
+    item['id'] = item['gwId'] = _get_gwid( itm )
```
jasonacox, line 1621: BUG: "item" should be "itm" - I hope that wasn't my bad influence. I'll commit a fix.

uzlonewolf:
Sigh.  I though I tested everything for sure before submitting this.  Python 2 and 3, wizard and scan, with and without \-force.  Of course I totally forgot about snapshot.  Again.  >_>

Thanks for the fix, it was a copypasta error.

Issue #266, opened 2023-01-29
Issues getting response from sensors


romicaiarca:
Hi,

I am trying to implement custom_component for Home Assistant but I have issues getting status from sensors. Here you have the exception that tiny tuya is returning on status call.

Requirement for tinytuya that I added in the component: ```"requirements": ["tinytuya>=1.10.0"],```

I have no idea why I receive this error. Is it possible to receive this because I instantiate 3 tintytuya objects for each parameter from sensor?

Thanks! 

```
Error doing job: Future exception was never retrieved
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/local_tuya_ths/THSTemperature.py", line 79, in get_data_from_sensor
    data = self.device.status()
  File "/usr/local/lib/python3.10/site-packages/tinytuya/core.py", line 1550, in status
    data = self._send_receive(payload, 0, getresponse=(not nowait))
  File "/usr/local/lib/python3.10/site-packages/tinytuya/core.py", line 912, in _send_receive
    return self.parent._send_receive(payload, minresponse, getresponse, decode_response, from_child=self)
  File "/usr/local/lib/python3.10/site-packages/tinytuya/core.py", line 1043, in _send_receive
    return self._process_message( msg, dev_type, from_child )
  File "/usr/local/lib/python3.10/site-packages/tinytuya/core.py", line 1101, in _process_message
    return self._send_receive( None, minresponse, True, decode_response, from_child=True)
NameError: name 'minresponse' is not defined
```

jasonacox:
Hi @romicaiarca - thanks for opening this issue!  This is a bug.  Both `decode_response` and `minresponse` are not in scope in that function.  I'll submit a fix.

@uzlonewolf  I believe this is all that is missing, but check my PR #267 .

```python
...
        return self._process_message( msg, dev_type, from_child )

    def _process_message( self, msg, dev_type=None, from_child=None ):
...
```
```python
...
        return self._process_message( msg, dev_type, from_child, minresponse, decode_response )

 def _process_message( self, msg, dev_type=None, from_child=None, minresponse=28, decode_response=True ):
...
```

jasonacox:
> I instantiate 3 tintytuya objects for each parameter

[...]

PR #267, opened 2023-01-29
Fix _process_message() missing params #266

jasonacox:
The `minresponse` and `decode_response` parameters are not being passed from _send_receive() to _process_message(). This adds those parameters.

### DIFF
--- a/tinytuya/core.py
+++ b/tinytuya/core.py
@@ -1040,9 +1040,9 @@ class XenonDevice(object):
             self._check_socket_close()
             return msg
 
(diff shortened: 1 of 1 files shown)


uzlonewolf:
Looks good to me.  Technically `minresponse` is no longer used and can be omitted, but there's no harm in having it.

jasonacox:
Thanks @uzlonewolf !  I'll rev us to 1.10.1 and add the BubDevice patch as well before releasing to PyPI.

PR #276, opened 2023-02-05
Added command line -debug flag and code cleanup based on pylint

jasonacox:
* Added `-debug` flag to command line interface to enable `set_debug()` for CL tools.
* Addressed all pylint warnings and errors or inserted `# pylint: disable=E0601` directives to ignore. There is a fairly generous .pylintrc exception list to ignore most style based warnings.


tinytuya/Contrib/ThermostatDevice.py
```diff
@@ -273,6 +273,7 @@ def __init__(self, *args, **kwargs):
         self.delayed_updates = { }
         self.sensorlists = [ ]
         self.sensors = self.SensorList( self )
+        self.mode = None
```
uzlonewolf, line 276: Is this to make pylint happy?  It won't hurt anything, but it's not really needed as it gets set via setattr() in the for loop below.

README.md
```diff
@@ -416,17 +416,19 @@ Tuya devices use AES encryption which is not available in the Python standard li
...
+python -m tinytuya <command> [<max_time>] [-debug] [-nocolor] [-force [192.168.0.0/24 192.168.1.0/24 ...]] [-h]
+
+  wizard         Launch Setup Wizard to get Tuya Local KEYs.
+  scan           Scan local network for Tuya devices.
+  devices        Scan all devices listed in devices.json file.
+  snapshot       Scan devices listed in snapshot.json file.
+  json           Scan devices listed in snapshot.json file [JSON].
+  <max_time>     Maximum time to find Tuya devices [Default=18]
+  -nocolor       Disable color text output.
+  -force         Force network scan for device IP addresses.  Auto-detects network range if none provided.
```
uzlonewolf, line 428: Would probably be a good idea to add a `    [net1/mask1 net2/mask2 ...]` line below this one

jasonacox: Agreed, I'll do that.

tinytuya/Contrib/ThermostatDevice.py
jasonacox: Great point and I almost dismissed it myself. At first I thought it might be a miss on our part, but then I saw it would work once setattr() is called in the flow which is why it wouldn't error.  I believe the basis for initializing all the classed based variables in __init__() is to help with readability.  I was confused at first before I traced it myself. Having the class scoped variables all listed in init() seems clean (maybe a bit pedantic too? :).

uzlonewolf: I hate duplicating stuff though, and listing everything would require duplicating the entire `dps_data` dict.  I just find it cleaner to have everything defined in 1 place instead of having smaller bits strewn about; this way there's no forgetting to initialize it in some other place after defining it here.

jasonacox: Fair. It's a style issue and not functional. Pylint (and vscode) are not able to unwind the setattr() to see that `self.mode` is actually defined in init().  but as you say, changing it would likely double the confusion and prevent someone from unwinding how you are doing the setter. I'll revert.

PR #285, opened 2023-02-17
Log message formatting

johnno1962:
Hi, was rumaging around trying to find out where my API quota was going and noticed messages were not having values inserted as perhaps you intended.

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -204,11 +204,11 @@ class Cloud(object):
         if action == 'GET':
             response = requests.get(url, headers=headers)
             log.debug(
-                "GET: response code=%d text=%s token=%s", response.status_code, response.text, self.token
+                "GET URL=%s HEADERS=%s: response code=%d text=%s token=%s" % (url, headers, response.status_code, response.text, self.token)
             )
         else:
             log.debug(
-                "POST: URL=%s HEADERS=%s DATA=%s", url, headers, body,
+                "POST: URL=%s HEADERS=%s DATA=%s" % (url, headers, body)
             )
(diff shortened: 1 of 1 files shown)


jasonacox:
Hi @johnno1962 - Thanks for opening this. 

Were you getting an error or the wrong data using the current version? There may be something else wrong.

I see the addition of `url` and `headers` here. This makes sense and would be a good addition! Thanks for that.

However, moving to the "%" format is not needed for debug().   In fact, pylint will flag this as suboptimal (see [W1201](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-not-lazy.html) ).  Using lazy logging lets the interpreter skip doing the text formatting until it is needed (if debug level was set). It's a small optimization but it is the reason you will see it in the code.

example test
```python
import logging

log = logging.getLogger(__name__)
logging.basicConfig(level=logging.DEBUG)

url = "url"
headers = "headers"
code = 200
text = "text"
token = "token"

log.debug("GET URL=%s HEADERS=%s: response code=%d text=%s token=%s" % (url, headers, code, text, token))

log.debug("GET URL=%s HEADERS=%s: response code=%d text=%s token=%s", url, headers, code, text, token)
```

johnno1962:
Ha, shows how much Python I know. So, the log function is performing the % expansion on demand which is as it should be. You can disregard this PR or, I've pushed another to commit to include only the URL=%s HEADERS=%s change. Thanks again for this invaluable resource. I've been able to talk to the cloud from Swift but exhausted my quota for this month :D

jasonacox:
Thanks @johnno1962 !  It's a good addition.

Thanks to this great community, and contributors like you, the project keeps getting better. ;)

PR #288, opened 2023-02-24
Add Cloud IR example, update docu, and allow an optional initial token to Cloud

### LINKED ISSUE #275: Add new function in class 'DEVICE'
Hi friend,
I am French, excuse me for my English ;)

I have equipment of the 'qn' category
Until now, I used the Cloud OpenAPI but it's a trial version that needs to be extended every 6 months... not great for home automation. 👎🏻 

While searching I came across your module '**tinytuya**' what happiness with local commands 🥇 

On the other hand, in my case, my equipment is a heater so when I send an ignition command, I also want to adjust the set temperature, the ventilation speed.

Rather than making 3 successive commands, I added a '**_set_value_dict()_**' function in the '**_Device_**' class in order to pass all the commands at once.

```
def set_value_dict(self, dict_value='', nowait=False):
        #Set value of any index.
        #Args:
        #    dict_value(dict): liste de commande index et value
        #    nowait(bool): True to send without waiting for response.

        # open device, send request, then close connection
        if (isinstance(dict_value, dict) and (len(dict_value) > 0)):
            payload = self.generate_payload(CONTROL, dict_value)
            data = self._send_receive(payload, getresponse=(not nowait))
        else:
            data["Err"]=914
            data["Error"]="dict_value required!"
        return data  
```

in my Python script :
```
        dict_reglages = {}
        dict_reglages["5"]= "2" # niveau de ventillation [1;2]
        dict_reglages["2"]= 29  # consigne de temperature [10;49]
        dict_reglages["1"]= False # etat du radiateur [True;False]
        # dict_reglages["1"]= True # etat du radiateur [True;False]
        
        # envoi la commande a l'equipement
        data = d.set_value_dict(dict_reglages)
        data = d.status() 
        print('%s ===> status() result %r' % (time.strftime('%H:%M:%S'),data))

# 12:28:59 ===> status(False 28 2) result {u'devId': u'102***********1a3c', u'dps': {u'1': False, u'2': 29, u'5': u'2'}, u't': 1675510137}
```

can you include it in your module?
**thank you very much for your work**

uzlonewolf:
The 1-2 second delay every time I called tinytuya.Cloud() was annoying me, so I added an option to pass in an initial auth token.  The existing code should auto-regenerate it if it has expired however I have not tested this.

Documentation for `d.set_multiple_values()` has been added to close #275 

A new example for Cloud IR sending has been added based upon the example I posted in #283

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -42,14 +42,12 @@ from .core import * # pylint: disable=W0401, W0614
 ########################################################
 
 class Cloud(object):
-    def __init__(self, apiRegion=None, apiKey=None, apiSecret=None, apiDeviceID=None, new_sign_algorithm=True):
+    def __init__(self, apiRegion=None, apiKey=None, apiSecret=None, apiDeviceID=None, new_sign_algorithm=True, initial_token=None):
         """
         Tuya Cloud IoT Platform Access
 
         Args:
-            dev_id (str): The device id.
-            address (str): The network address.
-            local_key (str, optional): The encryption key. Defaults to None.
+            initial_token: The auth token from a previous run.  It will be refreshed if it has expired
 
         Playload Construction - Header Data:
             Parameter 	  Type    Required	Description
@@ -81,7 +79,7 @@ class Cloud(object):
         self.apiDeviceID = apiDeviceID
         self.urlhost = ''
         self.uid = None     # Tuya Cloud User ID
-        self.token = None
+        self.token = initial_token
         self.error = None
         self.new_sign_algorithm = new_sign_algorithm
         self.server_time_offset = 0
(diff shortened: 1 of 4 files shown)


tinytuya/Cloud.py
```diff
@@ -466,6 +466,7 @@ def getconnectstatus(self, deviceid=None):
     def getdevicelog(self, deviceid=None, start=None, end=None, evtype=None, size=0, max_fetches=50, start_row_key=None, params=None):
         """
         Get the logs for a device.
+        https://developer.tuya.com/en/docs/cloud/0a30fc557f?id=Ka7kjybdo0jse
```
jasonacox, line 469: I both love including this and worry that they will change the URL over time. 🤷  Worth keeping.

tinytuya/core.py
```diff
@@ -26,6 +26,8 @@
     set_retry(retry=True)              # retry if response payload is truncated
     set_status(on, switch=1, nowait)   # Set status of switch to 'on' or 'off' (bool)
     set_value(index, value, nowait)    # Set int value of any index.
+    set_multiple_values(index_value_dict, nowait)
```
jasonacox, line 29: Good catch.

tinytuya/Cloud.py
uzlonewolf: I just figured you could always pull it out of the Internet Archive should it disappear.  https://web.archive.org/web/20230225125428/https://developer.tuya.com/en/docs/cloud/0a30fc557f?id=Ka7kjybdo0jse

PR #291, opened 2023-03-01
small tweaks to simplify

gstein:
Both floor() functions are unused, and math.ceil() is easily replaced.


uzlonewolf:
Very nice!  I like the `math` elimination.

Looks like that `floor()` function used to be used in the scanner before the rewrite.

jasonacox:
Thanks @gstein !

PR #300, opened 2023-03-14
Rework logging

uzlonewolf:
I recently discovered another issue with having tons of devices.  Like I mentioned in https://github.com/jasonacox/tinytuya/discussions/208#discussioncomment-4082687, I'm dumping a bunch of device sockets into select.select() just like the scanner does.  Do you know what happens if you turn debug on and then poll the status of 52 devices simultaneously?  It's... not pretty.  So, I reworked the logging.  Log messages now include the "name" of the source, which for devices is i.e. "tinytuya.OutletDevice.eb1######s".  Initially I removed the red coloring of the text from the message, but then decided it was probably there for a reason and put it back.  I did keep the device ID in cyan though.  Red removed:
Screenshot from 2023-03-13 20-14-56 cropped

Red put back:
Screenshot from 2023-03-13 23-55-49 cropped

When I added the per-device debug toggle I also made its color selection independent (default is to match the global).  I didn't intend it to be this way, but it worked out really well as it allows things like:
Screenshot from 2023-03-14 01-42-14 cropped

Anyway, I have this as a draft as it needs a \*lot\* more testing.  I've only verified it works for basic functions.

### DIFF
--- a/tinytuya/Contrib/ThermostatDevice.py
+++ b/tinytuya/Contrib/ThermostatDevice.py
@@ -215,7 +215,7 @@
 import struct
 import base64
 
-from ..core import Device, log, HEART_BEAT, DP_QUERY, CONTROL
+from ..core import Device, HEART_BEAT, DP_QUERY, CONTROL
 
 class ThermostatDevice(Device):
     """
@@ -379,7 +379,7 @@ class ThermostatDevice(Device):
                     break
 
         if not dps:
-            log.warn( 'Requested key %r not found!' % key )
+            self.log.warn( 'Requested key %r not found!' % key )
             return False
 
         ddata = self.dps_data[dps]
@@ -392,7 +392,7 @@ class ThermostatDevice(Device):
 
         if 'enum' in ddata:
             if val not in ddata['enum']:
-                log.warn( 'Requested value %r for key %r/%r not in enum list %r !  Setting anyway...' % (val, dps, key, ddata['enum']) )
+                self.log.warn( 'Requested value %r for key %r/%r not in enum list %r !  Setting anyway...' % (val, dps, key, ddata['enum']) )
 
         if 'base64' in ddata:
             val = base64.b64encode( val ).decode('ascii')
@@ -455,7 +455,7 @@ class ThermostatDevice(Device):
             for k in self.dps_data:
                 if k in data['dps'] and 'high_resolution' in self.dps_data[k]:
                     self.high_resolution = self.dps_data[k]['high_resolution']
-                    log.info('ThermostatDevice: high-resolution is now %r' % self.high_resolution)
+                    self.log.info('ThermostatDevice: high-resolution is now %r' % self.high_resolution)
                     break
 
         for k in data['dps']:
@@ -491,7 +491,7 @@ class ThermostatDevice(Device):
 
                     if 'enum' in self.dps_data[k]:
                         if val not in self.dps_data[k]['enum']:
-                            log.warn( 'Received value %r for key %r/%r not in enum list %r !  Perhaps enum list needs to be updated?' % (val, k, name, self.dps_data[k]['enum']) )
+                            self.log.warn( 'Received value %r for key %r/%r not in enum list %r !  Perhaps enum list needs to be updated?' % (val, k, name, self.dps_data[k]['enum']) )
 
                     if 'alt' in self.dps_data[k]:
                         data['changed'].append( self.dps_data[k]['alt'] )
@@ -616,7 +616,8 @@ class ThermostatDevice(Device):
                         coolto -= coolmod
                         if coolmod >= 25: coolto += 50
 
-                    log.info( 'CF is: %r %r %r cool: %r %r %r', cf, self.heatto, heatto / 100, self.coolto, coolto / 100, self.time )
+                    # FIXME: make a logger available here
+                    #log.info( 'CF is: %r %r %r cool: %r %r %r', cf, self.heatto, heatto / 100, self.coolto, coolto / 100, self.time )
 
                     # if self.time is a string then it needs to be in 24-hour HH:MM[:SS] format!
                     if isinstance( self.time, str ):
@@ -773,7 +774,7 @@ class ThermostatDevice(Device):
 
             if self.day_data[day][period][0] > 3 and self.day_data[day][period][1] < 1440:
                 if self.day_data[day][period][0] != 0xFF:
-                    log.warn('Selected participation flag is out of range, setting to %d', period)
+                    self.parent.log.warn('Selected participation flag is out of range, setting to %d', period)
                 self.day_data[day][period][0] = period & 3
 
         def setCF( self, cf ):
@@ -783,7 +784,7 @@ class ThermostatDevice(Device):
             self.have_data = False
 
             if len(data) % 7 != 0:
-                log.warn( 'Schedule data is in an unknown format, ignoring schedule' )
+                self.parent.log.warn( 'Schedule data is in an unknown format, ignoring schedule' )
                 return False
 
             cf = self.parent.getCF( self.cf )
@@ -793,7 +794,7 @@ class ThermostatDevice(Device):
                 day = data[offset:offset+daylen]
 
                 if len(day) % 7 != 0:
-                    log.warn( 'Schedule day data for day %d is in an unknown format, ignoring schedule' % dow )
+                    self.parent.log.warn( 'Schedule day data for day %d is in an unknown format, ignoring schedule' % dow )
                     return False
 
                 periods = len(day) / 7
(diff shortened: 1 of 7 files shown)


jasonacox:
I love this!!!

tinytuya/core.py
```diff
@@ -479,7 +488,6 @@ def parse_header(data):
         #seqno |= unknown << 32
         total_length = payload_len + header_len + len(SUFFIX_6699_BIN)
     else:
-        #log.debug('Header prefix wrong! %08X != %08X', prefix, PREFIX_VALUE)
```
jasonacox, line 482: While we are here, should we add a note?

```python
# Tuya protocol should include a header prefix 0x55AA (3.1, 3.2, 3.3, 3.4) or 0x6699 (3.5) 
# (see https://github.com/jasonacox/tinytuya/discussions/260)
```

And instead of "Header prefix wrong!" - should we error out with "Header prefix unknown!" ?  Not that Tuya would ever change that with the next protocol rev. 😉

tinytuya/Contrib/ThermostatDevice.py
```diff
@@ -616,7 +616,8 @@ def __bytes__( self ):
                         coolto -= coolmod
                         if coolmod >= 25: coolto += 50
 
-                    log.info( 'CF is: %r %r %r cool: %r %r %r', cf, self.heatto, heatto / 100, self.coolto, coolto / 100, self.time )
+                    # FIXME: make a logger available here
+                    #log.info( 'CF is: %r %r %r cool: %r %r %r', cf, self.heatto, heatto / 100, self.coolto, coolto / 100, self.time )
```
jasonacox, line 620: FIXME - I'm sure you know, just adding to my review

tinytuya/Contrib/ThermostatDevice.py
```diff
@@ -773,7 +774,7 @@ def setPeriod( self, day, period, **kwargs ):
 
             if self.day_data[day][period][0] > 3 and self.day_data[day][period][1] < 1440:
                 if self.day_data[day][period][0] != 0xFF:
-                    log.warn('Selected participation flag is out of range, setting to %d', period)
+                    self.parent.log.warn('Selected participation flag is out of range, setting to %d', period)
```
jasonacox, line 777: We are not inheriting or passing log to the child?

tinytuya/__main__.py
```diff
@@ -67,6 +66,18 @@
...
+#if debug:
+#    if color:
+#        logging.basicConfig( format="\x1b[0m\x1b[31;1m%(levelname)s:\x1b[22m%(name)s:\x1b[39;2m%(message)s\x1b[0m", level=logging.DEBUG )
+#    else:
+#        logging.basicConfig( level=logging.DEBUG )
+
+if state == 1:
+    from . import wizard
+else:
+    from . import scanner
```
jasonacox, line 79: I'm slow or tired or both... why do we want to defer?  Does scanner log logic infect wizard?

tinytuya/core.py
```diff
@@ -67,7 +67,7 @@
 import struct
 import sys
 import time
-from colorama import init
+from colorama import init, Fore, Back, Style
```
jasonacox, line 70: Ok, love this!!!

tinytuya/Contrib/ThermostatDevice.py
uzlonewolf: Yeah, I'm not sure what I'm going to do about that one, it's buried like 3 classes deep.  I may just remove it, though having a logger available there for debugging may be needed at some point.

tinytuya/core.py
```diff
@@ -333,13 +333,22 @@ def hex2bin(x):
 
 def set_debug(toggle=True, color=True):
     """Enable tinytuya verbose logging"""
+    log._tinytuya_use_color = color
+
+    if color:
+        # \x1b[39;2m
+        fmt = "\x1b[0m\x1b[31;1m%(levelname)s:\x1b[22m%(name)s:\x1b[31;1m%(message)s\x1b[0m"
```
jasonacox, line 340: I haven't tried, but wonder if we could leverage more colorama here instead to make it more readable. I mean, it's fine.

tinytuya/__main__.py
uzlonewolf: I'll most likely remove that one and put it back like it was.  I was doing a lot of experimenting and at one point tried to enable debugging before tinytuya was loaded to get the 2 version lines, but it ended up not working and I reverted most of it.

tinytuya/core.py
uzlonewolf: Yeah, I originally wanted to but couldn't decide between `a + b + '%(string)' + c +d` (ugh) or doing `%s%s%s%s%s%s` while trying to get the % formatting to not mangle the embedded %'s (also ugh).  I'll see if I can find a more-readable way that I'm happy with.

uzlonewolf: Ok, I think the choice is between
`fmt = ''.join( Style.RESET_ALL, Fore.RED, Style.BRIGHT, '%(levelname)s:', Style.NORMAL, '%(name)s:', Fore.RED, Style.BRIGHT, '%(message)s', Style.RESET_ALL )`
and
`fmt = Style.RESET_ALL + Fore.RED + Style.BRIGHT + '%(levelname)s:' + Style.NORMAL + '%(name)s:' + Fore.RED + Style.BRIGHT + '%(message)s' + Style.RESET_ALL`

Any preference?

jasonacox: I prefer the second one (addition).  Very simple.  Also, I can't get the first one to work in my test.

TypeError: str.join() takes exactly one argument (10 given)

uzlonewolf: Ah, I forgot the parentheses around the list, should have been `fmt = ''.join( (...) )`.  Oh well, we're using the additive one anyway.

While I was at it I rewrote `termcolor()`  as well:
```python
def termcolor(color=True):
    if color is False:
        # Disable Terminal Color Formatting
        bold = subbold = normal = dim = alert = alertdim = cyan = red = yellow = ""
    else:
        # Terminal Color Formatting
        bold     = Style.RESET_ALL + Fore.LIGHTWHITE_EX + Style.BRIGHT
        subbold  = Style.RESET_ALL + Fore.GREEN
        normal   = Style.RESET_ALL
        dim      = Style.RESET_ALL + Fore.LIGHTWHITE_EX + Style.DIM
        alert    = Style.RESET_ALL + Fore.LIGHTRED_EX + Style.BRIGHT
        alertdim = Style.RESET_ALL + Fore.LIGHTRED_EX + Style.DIM
        cyan     = Style.RESET_ALL + Fore.CYAN
        red      = Style.RESET_ALL + Fore.RED
        yellow   = Style.RESET_ALL + Fore.YELLOW
    return bold,subbold,normal,dim,alert,alertdim,cyan,red,yellow
```

jasonacox: For join, we would need to make it into an array.  I saw some post that join can be faster than addition, but it wasn't apparent to me and I still think the addition is cleaner:

```python
from colorama import Fore, Style
import time 

print("\nJoin Method: ")
start = time.time()
for x in range(200):
    fmt = ''.join([Style.RESET_ALL, Fore.RED, Style.BRIGHT, '%(levelname)s:', Style.NORMAL, '%(name)s:', Fore.RED, Style.BRIGHT, '%(message)s', Style.RESET_ALL])
end = time.time()
print("The time difference is :", end - start)

print("\nAddition Method: ")
start = time.time()
for x in range(200):
    fmt = (Style.RESET_ALL + Fore.RED + Style.BRIGHT + '%(levelname)s:' + Style.NORMAL + '%(name)s:' + Fore.RED + Style.BRIGHT + '%(message)s' + Style.RESET_ALL)
end = time.time()
print("The time difference is :", end - start)
```

tinytuya/Contrib/ThermostatDevice.py
jasonacox: Sortof gross, but possibly define a global scoped log (glog) to use for this one?

tinytuya/core.py
jasonacox: Nice! Yes, love the `termcolor()` update!

uzlonewolf: Upping the loop count to 200,000 makes it a bit more obvious, but the difference is so minuscule it's not worth worrying about.  I don't really care either way and currently have it as:
```python
    if color:
        fmt = Style.RESET_ALL + Fore.RED + Style.BRIGHT + '%(levelname)s:' + Style.NORMAL + '%(name)s:' + Fore.RED + Style.BRIGHT + '%(message)s' + Style.RESET_ALL
    else:
        fmt = "%(levelname)s:%(name)s:%(message)s"
```

PR #306, opened 2023-03-20
Fetch the device list a 2nd time to make sure we have the local key

### LINKED ISSUE #305: Wizard doesn't pull local_key for some devices
The hub that I just picked up, [eMylo 5.0 Bluetooth&3.0 ZigBee Gateway 3 In 1 ](https://www.amazon.com/dp/B09JNM1VW6?psc=1&ref=ppx_yo2ov_dt_b_product_details) (v3.4 device, Firmware: Main Module: v1.4.8 ZigBee Module: v2.4.1), has the same issue.  Wizard didn't pick up the local key (not in tuya-raw), but I was able to get it the "app" way as you mention.  I looked at the "result" raw from the `cloud.getdevices()` call and there is no local_key.

uzlonewolf:
Closes #305

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -213,6 +213,9 @@ class Cloud(object):
                 "POST: URL=%s HEADERS=%s DATA=%s", url, headers, body,
             )
             response = requests.post(url, headers=headers, data=body)
+            log.debug(
+                "POST RESPONSE: code=%d text=%s token=%s", response.status_code, response.text, self.token
+            )
 
         # Check to see if token is expired
         if "token invalid" in response.text:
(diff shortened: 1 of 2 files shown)


tinytuya/wizard.py
```diff
@@ -213,7 +226,10 @@ def wizard(color=True, retries=None, forcescan=False, nocloud=False):
             print('\n\n' + bold + 'Unable to save raw file' + dim )
 
     # Find out if we should poll all devices
-    answer = input(subbold + '\nPoll local devices? ' + normal + '(Y/n): ')
+    if quicklist:
```
jasonacox, line 229: Nice!

tinytuya/Cloud.py
```diff
@@ -18,7 +18,7 @@
         getfunctions(deviceid)
         getproperties(deviceid)
         getdps(deviceid)
-        sendcommand(deviceid, commands)
+        sendcommand(deviceid, commands [, uri])
```
jasonacox, line 21: Good catch

tinytuya/Cloud.py
```diff
@@ -364,6 +403,17 @@ def getdevices(self, verbose=False):
...
         else:
             json_data = self._get_all_devices()
+            users = {}
+            # loop through all devices and build a list of user IDs
+            for dev in json_data['result']:
+                if 'uid' in dev:
+                    users[dev['uid']] = True
+            if users:
+                # we have at least 1 user id, so fetch the device list again to make sure we have the local key
+                # this also gets us the gateway_id for child devices
```
jasonacox, line 413: Love this!

jasonacox:
Nice... I like the merge approach rather than scanning every ID with missing local_key's.

Testing... 

Do we want to try to get any of the draft changes in on this release?  I think this could go out as v1.12.1.

uzlonewolf:
Up to you.  Only reason I left #308 as a draft was because by the time I was done I was too tired to test it as much as I wanted.  #300 is going to require some slight updates after these 2 are merged.

jasonacox:
To keep it clean (and small batch changes), I released this PR as v1.12.1 (see https://pypi.org/project/tinytuya/1.12.1/).

PR #324, opened 2023-04-11
Cloud device list and Content-Type update

### LINKED ISSUE #323: Unable to find devices in wizard
Hi, I've been using tinytuya for a while but after upgrading to the latest version, I'm unable to find my devices using the wizard.  All of my details have stayed the same but it's returning 0 devices found (in the cloud).  If I run the local scan it finds all 8 devices.  I've double and triple checked my API key, Secret and example DeviceID but nothing seems to help.

Wondering if you might have any advice?

Many thanks,
Josh

uzlonewolf:
Changes:

The `Content-type: application/json` header is now always added (can be disabled by setting `_tuyaplatform( ..., content_type=None )`

If apiDeviceID is provided, the old device list fetching routine is used by default.  Leave apiDeviceID empty or set `cloud.use_old_device_list = False` to use the new routine.

apiDeviceID can now be a comma-separated list of device IDs and it uses all of them to build a list of UIDs.

The cloud init function now ignores any extra/unknown arguments, so you can have extra fields in tinytuya.json and `tinytuya.Cloud( **config )` will still work.

Fixes #323

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -42,7 +42,7 @@ from .core import * # pylint: disable=W0401, W0614
 ########################################################
 
 class Cloud(object):
-    def __init__(self, apiRegion=None, apiKey=None, apiSecret=None, apiDeviceID=None, new_sign_algorithm=True, initial_token=None):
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks @uzlonewolf !  I'll push this as v1.12.3.

PR #326, opened 2023-04-11
Fix fan_run_time typo in Contrib/ThermostatDevice

elockman:
This is needed to update the fan runtime during fan cycle mode

### DIFF
--- a/tinytuya/Contrib/ThermostatDevice.py
+++ b/tinytuya/Contrib/ThermostatDevice.py
@@ -347,7 +347,7 @@ class ThermostatDevice(Device):
         return self.setValue( 'hold', hold )
 
     def setFanRuntime( self, rt ):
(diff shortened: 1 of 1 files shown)


elockman:
Resolves the following error:
```
Requested key 'fan_runtime' not found!
  File "/home/user/projects/xeal/tuya/new_test/tinytuya/examples/test_stat.py", line 31, in <module>
    tstatdev.setFanRuntime( 15 )
  File "/home/user/.local/lib/python3.9/site-packages/tinytuya/Contrib/ThermostatDevice.py", line 350, in setFanRuntime
    return self.setValue( 'fan_runtime', int(rt) )
  File "/home/user/.local/lib/python3.9/site-packages/tinytuya/Contrib/ThermostatDevice.py", line 353, in setValue
    dps, val = self.parseValue( key, val )
TypeError: cannot unpack non-iterable bool object
```

PR #308, opened 2023-03-20
Gateway/sub-device updates, payload_dict reworked, and subdev_query() added

uzlonewolf:
Quite a few changes to payload_dict processing.  I don't expect it to affect non-gateway 3.1-3.3 devices at all, but for v3.4+ I changed the "protocol" key from 5 to 4.  My packet captures seem to show the app and devices use them interchangeably and I have not noticed any issues with "4" on my v3.4 devices.

Said packet captures also showed the app polling for sub-device status, so I implemented that via the new `subdev_query()` function.  My v3.3 gateways just ignore it though, so it's probably a v3.4+-only thing.
Request: cmd=64, `'{"data":{"cids":[]},"reqType":"subdev_online_stat_query"}`
Response: `{"reqType":"subdev_online_stat_report","data":{"online":["tu###4"],"offline":[],"nearby":["tu###4"]}}`
I poked around in the firmware a bit but didn't see any other "reqType" commands which actually did anything.

Closes #302 
Possibly closes #266 
Closes #307 
Closes #309

PR #331, opened 2023-04-16
Allow wizard to scan for a Device ID to use

uzlonewolf:
Screenshot from 2023-04-16 12-55-17

### DIFF
--- a/tinytuya/wizard.py
+++ b/tinytuya/wizard.py
@@ -138,7 +138,7 @@ def wizard(color=True, retries=None, forcescan=False, nocloud=False, quicklist=F
         with open(DEVICEFILE, "r") as infile:
             tuyadevices = json.load( infile )
     else:
-        if 'apiDeviceID' in config and config['apiDeviceID'] and config['apiDeviceID'].trim().lower() == 'scan':
+        if 'apiDeviceID' in config and config['apiDeviceID'] and config['apiDeviceID'].strip().lower() == 'scan':
             config['apiDeviceID'] = ''
             print( '\nScanning to find a Device ID...' )
             dev = tinytuya.scanner.devices( verbose=False, poll=False, byID=True, show_timer=False, maxdevices=1 )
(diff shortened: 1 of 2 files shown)


uzlonewolf:
With the "associated-users" API endpoint getting blocked I can see an argument where an empty DeviceID should also kick off a scan, but I do hope we can eventually find another way to get the device list without needing a device ID.

jasonacox:
I wonder why they would restrict this? My only thought is that there could be an attack vector they are trying to harden a bit (e.g. bad actor trying to fetch all of your devices after somehow gaining access somehow to your IoT API credentials).

tinytuya/wizard.py
```diff
@@ -138,6 +138,18 @@ def wizard(color=True, retries=None, forcescan=False, nocloud=False, quicklist=F
         with open(DEVICEFILE, "r") as infile:
             tuyadevices = json.load( infile )
     else:
+        if 'apiDeviceID' in config and config['apiDeviceID'] == 'scan':
```
jasonacox, line 141: Should we make that `config['apiDeviceID'].lower() == 'scan':` to ignore user entered caps?

uzlonewolf: Done!

uzlonewolf:
I have no idea.

I was looking through the "official" Python API implementation at https://github.com/tuya/tuya-iot-python-sdk/blob/main/example/api.py and they get the device list by calling `/v1.0/iot-03/users/assets` followed by `/v1.0/iot-02/assets/{}/devices`, but of course they return "permission deny" as well.  Looking at the "Quick Start Guide" at https://developer.tuya.com/en/docs/iot/quick-start1?id=K95ztz9u9t89n I wonder if there's an API Service we need to add authorization for to get access to these.

uzlonewolf:
Going a bit off topic here, but while looking through that repository I noticed `PROTOCOL_DEVICE_REPORT = 4` and thought "hmmm, 'protocol: 4' looks familiar."  A bit more digging reveals while 4 and 20 are the only protocols defined there, https://github.com/tuya/tuya-connect-kit-for-mqtt-embedded-c/blob/without-submodules/include/mqtt_service.h lists a bunch more.  So, it turns out `protocol: 4` means `dev -> cloud push dp data` and `protocol: 5` is `cloud -> dev send dp data`.  Mystery from #308 solved!

jasonacox:
Wow! Amazing find!!  I wonder how we could capture this info gold?  Maybe add it to https://github.com/jasonacox/tinytuya/discussions/260 ?

PR merged. Love the scan function!  I'll push this as v1.12.4 as a minor patch.

Issue #333, opened 2023-04-19
Typo or intentional?


teejo75:
https://github.com/jasonacox/tinytuya/blob/de222a5bae96b0ef1bf0c5babf0d6e79648f13ce/tinytuya/scanner.py#L71

Is that supposed to be `ablilty`?

uzlonewolf:
That's what devices report.  It's a typo on Tuya's part, just like boardcast.

teejo75:
Thanks for the insight. I searched the codebase and there were no code hits for 'ability' and only scanner.py for the typo. I was curious.

PR #336, opened 2023-04-23
Misc minor updates

uzlonewolf:
A few random minor updates.

1. Normalize ability/ablilty in scanner #333 
2. Fix examples/getstatus.py (`d.get_status()` does not exist)
3. Partially revert #324.  I don't think it's causing problems, but setting Content-type on non-POST requests is rubbing me the wrong way.
4. Rework examples/async_send_receive.py

### DIFF
--- a/examples/getstatus.py
+++ b/examples/getstatus.py
@@ -26,5 +26,5 @@ d.set_version(3.3)
 
 # Get the status of the device 
 # e.g. {'devId': '0071299988f9376255b', 'dps': {'1': True, '3': 208, '101': False}}
-data = d.get_status()
+data = d.status()
 print(data)
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -127,7 +127,7 @@ class Cloud(object):
         if self.apiRegion == "in":
             self.urlhost = "openapi.tuyain.com"      # India Datacenter
 
-    def _tuyaplatform(self, uri, action='GET', post=None, ver='v1.0', recursive=False, query=None, content_type='application/json'):
+    def _tuyaplatform(self, uri, action='GET', post=None, ver='v1.0', recursive=False, query=None, content_type=None):
         """
         Handle GET and POST requests to Tuya Cloud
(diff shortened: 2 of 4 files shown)


tinytuya/Cloud.py
```diff
@@ -127,7 +127,7 @@ def setregion(self, apiRegion=None):
         if self.apiRegion == "in":
             self.urlhost = "openapi.tuyain.com"      # India Datacenter
 
-    def _tuyaplatform(self, uri, action='GET', post=None, ver='v1.0', recursive=False, query=None, content_type='application/json'):
+    def _tuyaplatform(self, uri, action='GET', post=None, ver='v1.0', recursive=False, query=None, content_type=None):
```
jasonacox, line 130: Agree with this. It could be considered a breaking change for the API, but since this is a _function() we aren't breaking the contract.

tinytuya/scanner.py
```diff
@@ -170,6 +170,10 @@ def __init__( self, ip, deviceinfo, options, debug ):
 
         if not deviceinfo:
             deviceinfo = {}
+        # some devices report "ability" but most have this as the typo "ablilty"
+        if 'ablilty' in deviceinfo and 'ability' not in deviceinfo:
+            deviceinfo['ability'] = deviceinfo['ablilty']
+            del deviceinfo['ablilty']
```
jasonacox, line 176: Spell check for Tuya. :)

examples/getstatus.py
```diff
@@ -26,5 +26,5 @@
 
 # Get the status of the device 
 # e.g. {'devId': '0071299988f9376255b', 'dps': {'1': True, '3': 208, '101': False}}
-data = d.get_status()
-print(data)
\ No newline at end of file
+data = d.status()
+print(data)
```
jasonacox, line 30: wow, how long has that been there? :)

jasonacox:
Thanks @uzlonewolf !

Added this into v1.12.5.  Released.

PR #228, opened 2022-12-02
Add support for newer IR devices, and several IR format converters

uzlonewolf:
Also adds `set_multiple_values()` to core as the newer devices require everything to be in one packet.  I also updated turn_on() and turn_off() to return the received data instead of just dropping it on the floor while I was in there.

Documentation for new format converters and head/key sending in examples/Contrib/IRRemoteControlDevice-example.py

Closes #225 and I'm also going to say Closes #74 and Closes #27

PR #343, opened 2023-05-06
Add Contrib support for IR+RF devices such as the S11

uzlonewolf:
This adds Contrib/RFRemoteControlDevice as a subclass of Contrib/IRRemoteControlDevice.  See discussion in #228 

I also removed some wayward tabs from IRRemoteControlDevice and ThermostatDevice.

### DIFF
--- a/examples/Contrib/RFRemoteControlDevice-example.py
+++ b/examples/Contrib/RFRemoteControlDevice-example.py
@@ -1,0 +1,20 @@
+#import tinytuya
+#tinytuya.set_debug()
+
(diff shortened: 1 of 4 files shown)


jasonacox:
Thanks @uzlonewolf !  

Also nice cleanup on tab/spaces. I'm wondering if we should add something like .editorconfig to the project - see https://EditorConfig.org

jasonacox:
https://pypi.org/project/tinytuya/1.12.6/

PR #345, opened 2023-05-10
fix: call status() reset device

xgustavoh:
Hello, first I would like to apologize for my English!

Recently I had a problem using a **Smart Bulb RGBCW7** in `version = 3.2`, that it was turning off or changing its status "randomly".

I noticed that whenever the `status()` function was executed, the bulb suffered a "reset" or "turning off".

After seeing the code, I noticed that the `dps_cache` was exactly to the value that the bulb was getting after the "reset".

When I set the `dps_to_request` values to `None, the problem didn't happen.

---

Adding this code in `def detect_available_dps(self):` solved my problem.
```python
data["dps"] = {k: None for k in data["dps"]}
```

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,10 @@
 # RELEASE NOTES
 
+## v1.12.7 - Status Bug Fix
+
+* PyPI 1.12.7
+* Fix bug in `detect_available_dps()` to resolve issue where `status()` call for smartbulbs would randomly cause devices to turn off by @xgustavoh in https://github.com/jasonacox/tinytuya/pull/345
+
 ## v1.12.6 - Minor Fixes
 
 * PyPI 1.12.6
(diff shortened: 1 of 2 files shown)


jasonacox:
Thanks @xgustavoh !  I'll run some tests to verify

@uzlonewolf can you take a look as well?

uzlonewolf:
I like the idea, though I think creating a new dict just to update another is kinda convoluted.  How about replacing all of that with:

```python
            if data is not None and "dps" in data:
                # save the keys but not the returned values
                for k in data["dps"]:
                    self.dps_cache[k] = None
```

xgustavoh:
I like your way too, it even had implementation that way...
But, I'm not a python programmer, so I don't know the best practices, I'm going to change right now!

jasonacox:
Thank you @xgustavoh !!   My tests are passing. I'll release it as v1.12.7.

PR #352, opened 2023-05-21
Use {DeviceName} instead of {DeviceID} alternatively for API commands

mschlenstedt:
I added the possibility to use {DeviceName} instead of {DeviceID} in API commands for a more convinient way to set up API calls. If one uses {DeviceName} the Device Names must be absolutely unique, of course.

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -41,6 +41,7 @@ import os
 import urllib.parse
 from http.server import BaseHTTPRequestHandler, HTTPServer, ThreadingHTTPServer
 from socketserver import ThreadingMixIn 
+import urllib.parse
 
 # Required module: pycryptodome
 try:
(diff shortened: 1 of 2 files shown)


server/server.py
```diff
@@ -384,6 +385,11 @@ def do_GET(self):
             except:
                 message = json.dumps({"Error": "Syntax error in set command URL.", "url": self.path})
                 log.debug("Syntax error in set command URL: %s" % self.path)
+            if(id not in deviceslist):
+                for key in deviceslist:
+                    if deviceslist[key]['name'] == urllib.parse.unquote(id):
+                        id = deviceslist[key]['id']
+                        break
```
jasonacox, line 392: Minor note: Since we are repeating this same logic 6 times, should we consider making this a function?

mschlenstedt: Yes, of course. I was unsure to put it into a function... I will change it accordingly.

uzlonewolf: I would also pull the `urllib.parse.unquote(id)` out of the loop and move it to a variable beforehand instead.

mschlenstedt: I included both suggestions: getDeviceIdByName as a function and pull `urllib.parse.unquote(id)` out of the loop.

jasonacox:
Thanks, @mschlenstedt ! Nice addition.

```bash
# by DeviceID
$ curl http://localhost:8888/status/xxxxxxxxxxxxxxxxxx01
{"devId": "xxxxxxxxxxxxxxxxxx01", "dps": {"1": true, "9": 0}}

# by DeviceName
$ curl http://localhost:8888/status/Kitchen%20Light
{"devId": "xxxxxxxxxxxxxxxxxx01", "dps": {"1": true, "9": 0}}
$ curl http://localhost:8888/status/SmartBulb                                
{"devId": "xxxxxxxxxxxxxxxxxx02", "dps": {"20": true, "21": "white", "22": 1000, "24": "000003e803e8", "25":"07464602000003e803e800000000464602007803e803e80000000046460200f003e803e800000000464602003d03e803e80000000046460200ae03e803e800000000464602011303e803e800000000", "26": 0}}
```

I updated the build to `t7` and will push it to dockerhub.

PR #353, opened 2023-05-23
Mapping for DP IDs

mschlenstedt:
Hi all,

I added a first "Proof of Concept" for doing a DPS <-> Name/Code mapping to have a more comfortable way to set and read the DPS.

At a first step I included mappings from IOBroker and FHEM project (they already use the DPS<->Code/Name mapping). The mapping is read once at startup after reading the devices.json and after saving a new deivces.json. Only those mappings are kept in memory (in a dict) which are currently used. This keeps things small and almost fast.

`/status/{DEVICE_ID}` now additionally give back the mapping scheme if one exists for the Product_Key in `mappings/mappings.json`:

```
{"devId": "05047xxxxxx923ca8", "dps": {"1": false, "2": 21, "3": 13, "4": "hot", "9": 256}, "dps_scheme": [{"id": 1, "code": "Power"}, {"id": 2, "code": "temp_set"}, {"id": 3, "code": "temp_current"}, {"id": 4, "code": "mode"}, {"id": 9, "code": "fault"}]}
```

At a first step I included the mapping in the WebUI:

image

What's still missing:
- Re-Read mappings if a device is discovered
- Add mappings in `/set/....` command

PR #356, opened 2023-05-24
New Cloud functions to download DP Name mappings

uzlonewolf:
Adds `mappings.json` to hold the DP<->Name mappings and various functions to download, save, and load it.  `mappings.json` contains a dict with product IDs as keys.

Submitting as a draft as I still need to add documentation but wanted to show the direction I was going in after the discussion in #353.

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -84,6 +84,7 @@ class Cloud(object):
         self.new_sign_algorithm = new_sign_algorithm
         self.server_time_offset = 0
         self.use_old_device_list = True
+        self.mappings = None
 
         if (not apiKey) or (not apiSecret):
             try:
@@ -737,3 +738,71 @@ class Cloud(object):
         elif len(str(ts)) == 10:
             ts *= 1000
         return ts
+
+    @staticmethod
+    def _build_mapping( src, dst ):
+        for mapp in src:
+            try:
+                code = mapp['code']
+                dp_id = str(mapp['dp_id'])
+                if dp_id in dst:
+                    continue
+                data = { 'name': code, 'type': mapp['type'] }
+                try:
+                    values = json.loads( mapp['values'] )
+                except:
+                    values = {}
+                if values and 'unit' in values:
+                    if values['unit']:
+                        values['unit'] = values['unit'].replace('℉','°F').replace('℃','°C').replace('f','°F').replace('c','°C').replace('秒','s') #.replace('份','')
+                data['values'] = values
+                dst[dp_id] = data
+            except:
+                log.debug( 'Parse mapping item failed!', exc_info=True )
+
+    def getmapping( self, productid, deviceid=None ):
+        if not self.mappings:
+            self.mappings = {} #load_mappings()
+
+        if productid in self.mappings:
+            return self.mappings[productid]
+
+        if deviceid:
+            result = self.getdps(deviceid)
+            if result:
+                if 'result' in result:
+                    result = result['result']
+                    dps = {}
+                    if 'status' in result:
+                        self._build_mapping( result['status'], dps )
+                    if 'functions' in result:
+                        self._build_mapping( result['functions'], dps )
+                    self.mappings[productid] = dps
+                    log.debug( 'Loaded mapping for device %r: $r', deviceid, dps)
+                elif ('code' in result and result['code'] == 2009) or ('msg' in result and result['msg'] == 'not support this device'):
+                    self.mappings[productid] = {}
+
+        if productid in self.mappings:
+            return self.mappings[productid]
+
+        return None
+
+    def setmappings( self, mappings ):
+        if type(mappings) == dict:
+            self.mappings = mappings
+
+    def getmappings( self, devices ):
+        if not self.mappings:
+            self.mappings = {}
+
+        for dev in devices:
+            try:
+                devid = dev['id']
+                productid = dev['product_id']
+            except:
+                continue
+
+            if productid not in self.mappings:
+                self.getmapping( productid, devid )
+
+        return self.mappings
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -741,57 +741,76 @@ class Cloud(object):
 
     @staticmethod
     def _build_mapping( src, dst ):
+        # merge multiple DPS sets from result['status'] and result['functions'] into a single result
         for mapp in src:
             try:
                 code = mapp['code']
                 dp_id = str(mapp['dp_id'])
                 if dp_id in dst:
                     continue
-                data = { 'name': code, 'type': mapp['type'] }
+                data = { 'code': code, 'type': mapp['type'] }
                 try:
                     values = json.loads( mapp['values'] )
                 except:
                     values = {}
                 if values and 'unit' in values:
                     if values['unit']:
-                        values['unit'] = values['unit'].replace('℉','°F').replace('℃','°C').replace('f','°F').replace('c','°C').replace('秒','s') #.replace('份','')
+                        # normalize some unicode and temperature units
+                        # not sure what a good replacement for '份' is (seen in a pet feeder)
+                        values['unit'] = values['unit'].replace('℉','°F').replace('℃','°C').replace('f','°F').replace('c','°C').replace('秒','s')
+
+                # Tuya's 'JSON' mapping type is an ordered dict, but python's dict is not!  so, save the raw value as well
+                if mapp['type'].lower() == 'json':
+                    data['raw_values'] = mapp['values']
                 data['values'] = values
                 dst[dp_id] = data
             except:
                 log.debug( 'Parse mapping item failed!', exc_info=True )
 
     def getmapping( self, productid, deviceid=None ):
+        # return a mapping for the given product id, or download it from the cloud using a device id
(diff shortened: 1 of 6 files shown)


mschlenstedt:
So if I understood your changes correctly, the wizard will create the mappings.json? So I do not have to handle this in PF #353 ? I just call "getmappings( self, devices )" and that's it?

mschlenstedt:
Ah, no, I have to call "load_mappings(mappings)", right?

uzlonewolf:
If you only have 1 or 2 devices then you can use the new `find_mapping(product_id or device_id)` function I just added.  Otherwise, if you have more, I recommend something along the lines of:

```python
import tinytuya
import json

mappings = tinytuya.load_mappings()

with open(tinytuya.DEVICEFILE, "r") as infile:
    tuyadevices = json.load( infile )

for dev in tuyadevices:
    try:
        devid = dev['id']
        productid = dev['product_id']
    except:
        # we need both the device id and the product id to download mappings!
        print( 'No device id and/or product id!', dev )
        continue

    if productid in mappings:
        print( 'Mapping for device %s:' % devid, mappings[productid] )
    else:
        print( 'Device %s has no mappings!' % devid )
```

uzlonewolf:
With that, I think this first half is done.  I have the "download mappings?" question in the wizard defaulting to "no" for now, not sure if I should change that to "yes"

uzlonewolf:
Ok, I added another function to assign mappings to an entire tuyadevice list.  Now all you need to do is:

```python
import tinytuya
import json

with open(tinytuya.DEVICEFILE, "r") as infile:
    tuyadevices = json.load( infile )

print( 'before:', tuyadevices )
tinytuya.assign_dp_mappings( tuyadevices )
print( 'after:', tuyadevices )
```

mschlenstedt:
> I have the "download mappings?" question in the wizard defaulting to "no" for now, not sure if I should change that to "yes"

I would change it to yes. Not much overload.

tinytuya/wizard.py
```diff
@@ -237,6 +237,18 @@ def wizard(color=True, retries=None, forcescan=False, nocloud=False, quicklist=F
         except:
             print('\n\n' + bold + 'Unable to save raw file' + dim )
 
+    # Fetch the DP name mappings for all devices
+    if (not nocloud) and (not quicklist):
+        # Default to 'no'
+        answer = input(subbold + '\nDownload DP Name mappings? ' + normal + '(y/N): ')
+        if answer.lower().find('y') >= 0:
```
jasonacox, line 244: Agree we should default to Y.

tinytuya/core.py
```diff
@@ -109,6 +124,7 @@
 DEVICEFILE = 'devices.json'
 RAWFILE = 'tuya-raw.json'
 SNAPSHOTFILE = 'snapshot.json'
+DPMAPPINGSFILE = 'mappings.json'
```
jasonacox, line 127: Is there a reason we want this as a separate file instead of embedding the map directly into devices.json?  I'm fine with as-designed, just making sure we aren't missing something.

uzlonewolf: We definitely can, I just had initially approached this as an optional add-on and not a core part.  If we're going to default to 'yes' then I think it would make more sense to just roll this into the device list downloader instead of calling it as a separate step.

uzlonewolf: Though having it as a separate file does allow us to skip downloading mappings we already have.  Not sure if this is good or bad...  Good in that it saves API calls, bad in that we do not get the latest data every time.  Personally it saves me 53 calls every time I run the wizard.  Perhaps it would be better to download it every time and create a new "only fetch newly added devices" option for the wizard.

jasonacox: I like that idea, a lot.  It add some complexity but also helps avoid too many API calls.  Good call out on that. We have the tension between data and limited Tuya cloud API calls we need walk.

My thought on "default to yes" is that it lets us build on the assertion that the `devices.json` file (if we choose to put it there) can contain this mapping data and be used by other apps (like the `server.py` example by @mschlenstedt , and also other home automation tools/dashboards that already use that file).  Those tools can magically unlock new capabilities because it is there. If we default to "no", I'm going to assume most new users will select the default so expanded device data (mapping) would be rare. I know, it is subtle but I do think it promotes the expansion and use of the additional attributes. For the existing TinyTuya user base, when they upgrade to the new version and run wizard, they will get a prompted default to add the mapping data as well.

What do you both think?  I see you switched us back to draft @uzlonewolf.  Are you wanting to take a try at this?

uzlonewolf: I agree with everything said so far.  Yes, I switched back to draft so I could make these changes.

After ripping out DPMAPPINGSFILE and merging everything into devices.json, my devices.json went from 64 KiB / 1851 lines to 454 KiB / 13871 lines.  I do have a ton of devices though (127 to be exact) so this may be acceptable, not sure.  For comparison, my mappings.json was 115 KiB and 4935 lines.

uzlonewolf:
After ripping out DPMAPPINGSFILE and merging everything into devices.json, my devices.json went from 64 KiB / 1851 lines to 454 KiB / 13871 lines. I do have a ton of devices though (127 to be exact) so this may be acceptable, not sure. For comparison, my mappings.json was 115 KiB and 4935 lines.

tinytuya/core.py
jasonacox: Wow! And thought I was eccentric with my 41 devices. 😄 

TBH, I don't have any problem with a 454k device file for 127 devices.  Most users are likely 10% of that but even at 4k/device, that seems reasonable for the additional functionality.  Is the difference between mappings.json due to redundant definitions (multiple same product types)?

uzlonewolf:
Also, Cloud.getdevices() now skips downloading the factory-infos and DP mappings unless something (such as the local key) has changed.  For my massive device list this drops the wizard run time from 34 seconds to ~5 in addition to saving multiple dozens of API calls.

tinytuya/core.py
uzlonewolf: Yep, 127 total devices with 53 different product id's, so instead of 53 mappings there's now 127 mappings.  I have a boatload of same-model smart bulbs in track lights...

tinytuya/Cloud.py
```diff
@@ -451,9 +458,55 @@ def getdevices(self, verbose=False):
...
+            old = old_devices[dev_id]
+            if 'key' not in old or old['key'] != dev['local_key']:
+                # local key changed
+                changed_devices.append( dev )
+                continue
+            if (('icon' not in old) and ('icon' in dev)) or (include_map and ('mapping' not in old or old['mapping'] is None)):
+                # icon or mapping added
+                #changed_devices.append( dev )
+                #continue
+                pass
```
jasonacox, line 483: Is this on purpose?

If you have an existing `devices.json` (most existing users), it will not download the DP mappings since it finds the device already there.  Suggestion we consider:

* If user requests to get DP mappings, and DEVICEFILE exists, prompt if they want to replace file or only add new devices
* Adjust this section of `getdevices()` to add mappings if missing (currently `pass`)?

uzlonewolf: Oops, yeah, that was commented out for testing and I meant to enable it before pushing.  Fixed.

jasonacox: Awesome!  

FYI, for my 41 devices `devices.json` was 23k before and now 91k.  

This data is amazing, @uzlonewolf.  Nice job.  Merging and testing....

jasonacox:
Tested - no issues - released: https://pypi.org/project/tinytuya/1.12.8/

Thanks again @uzlonewolf - you're brilliant! 🙏 

@mschlenstedt, the new `devices.json` should support the DP mappings for #353 now.

mschlenstedt:
Yes, I will change my PR to use the new file. Will take a few days though.

> If user requests to get DP mappings, and DEVICEFILE exists, prompt if they want to replace file or only add new devices

Haven't checked already, but I would prefer that the default ("yes") recreates the devices.json in total. I run the wizard from my Frontend via script this way:

`yes | python3 -m tinytuya wizard`

So everything what is not "yes" will be missed... An even better way of course would be to define the answers e.g. in ENVIRONMENT variables. But this is maybe another PR...

jasonacox:
No rush! Thanks for your PR and engagement on this enhancement.

> Haven't checked already, but I would prefer that the default ("yes") recreates the devices.json in total. I run the wizard from my Frontend via script this way:
>
>yes | python3 -m tinytuya wizard

This works for me. The last commit @uzlonewolf made doesn't need to recreate devices.json, it enables the logic to add mappings to the device data if they are missing.  I confirmed by restoring an old devices.json and running the wizard and saw that it added DP mappings to all devices.

uzlonewolf:
> An even better way of course would be to define the answers e.g. in ENVIRONMENT variables. But this is maybe another PR...

Actually this PR also abuses the scanner's `-y`/`-yes` flag to answer "yes" to the "Use existing credentials?" and "Download DP Name mappings?" questions but "no" to "Poll local devices?"  I had added an argument to `wizard()` a long time ago to do this, and after the timing tests I did for this PR I just left the command line switches in.  Just call with `python3 -m tinytuya wizard -yes`.  As a future PR it would probably be a good idea to change this to something more intuitive (perhaps `-yes-cloud` or something?) and add it to the help documentation.

PR #363, opened 2023-06-11
Doubled import and Typo

mschlenstedt:
Just two small fixes:

- urllib.parse was imported twice
- This is the API Server not the Wizard ;-)


jasonacox:
Thanks @mschlenstedt !  Good catch. :)

PR #364, opened 2023-06-11
MQTT Gateway for Server

mschlenstedt:
Hey,

I really love the idea to include MQTT into the server. I created a seperate gateway script which passes status of devices from the Server to a MQTT Broker and subscribes to a "set" topic and passes the payload back to the set API call of the Server.

The script polls the status of all known devices (in seperate threads, 20 at the same time currently) every x seconds. Polling is not the best solution here of course, I would really love to have a persist socket to each device and get status changes immediately. But I think this needs a complete rewrite of the server. So maybe polling is "better than nothing" here.

Code needs definetly some clean up and documentation is still missing.

### DIFF
--- a/server/mqtt/mqtt.json
+++ b/server/mqtt/mqtt.json
@@ -1,0 +1,8 @@
+{
+   "broker" : "localhost",
+   "port" : 1883,
+   "username" : "Broker_Username",
+   "password" : "Broker_Password",
+   "topic" : "tinytuya",
+   "pollingtime" : 5
+}
--- a/server/mqtt/mqtt_gateway.py
+++ b/server/mqtt/mqtt_gateway.py
@@ -1,0 +1,217 @@
+# TinyTuya MQTT Gateway
+# -*- coding: utf-8 -*-
+"""
+TinyTuya MQTT Gateway for API Server for Tuya based WiFi smart devices
+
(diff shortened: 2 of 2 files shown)


mschlenstedt:
MQTT topic structure:

image

uzlonewolf:
So, I'm currently in the middle of a server rewrite that does all this and more.  It currently makes persistent connections to all devices and does this with a single thread via non-blocking sockets, supports HTTP, WebSockets, plain TCP, and all of the above wrapped in SSL/TLS.  It also has hooks for native MQTT but I have not yet had a chance to hook them up yet https://github.com/uzlonewolf/tinytuya/blob/socket-server/multiserver/multiserver.py#L963 .  I'm currently out of town and have not had much time to work on it, but should be able to get back to it next week.

mschlenstedt:
Fully agree - best option is to have persist socket connections. Maybe it's a good idea to include the mqtt_gateway into the existing server until your new rewrite is ready for production. If you leave the MQTT structure the same, users can easily upgrade.

jasonacox:
Thanks @mschlenstedt !  I agee, this is a good incremental feature for our current server.  

I know @uzlonewolf 's multiserver is going to be amazing!  Safe travel, LW!

PR #366, opened 2023-06-17
Do not use hardcoded port number

mschlenstedt:
Removed hardcoded port number - in case someone would like to configure it (change it from 8888).


jasonacox:
Good fix @mschlenstedt !  Thank you.

PR #367, opened 2023-06-17
Avoid exception if username/pwassword not defined or empty

mschlenstedt:


mschlenstedt:
Also avoid exception if data type of MQTT Polling time is not set correctly.

Noschvie (approved):

jasonacox:
Thanks @mschlenstedt and @Noschvie !

PR #368, opened 2023-06-18
Add Contrib support for Inverter Heat Pump such as Fairland IPHR55

valentindusollier:
This adds support for Inverter Heat Pump such as Fairland Inverter+ 21kW (IPHR55).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,3 @@
 pycryptodome        # Encryption - AES can also be provided via PyCrypto or pyaes
 requests            # Used for Setup Wizard - Tuya IoT Platform calls
 colorama            # Makes ANSI escape character sequences work under MS Windows.
-python-dotenv       # Reads .env file for environment variables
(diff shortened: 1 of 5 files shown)


valentindusollier:
I made the dotenv module optional. It is only use in `/testcontrib.py` in order to load environment variables from .env file. However it's not really needed. Moreover, this module is not installed in the workflows.

uzlonewolf:
I do like the idea, but unfortunately python-dotenv only supports Python v3.8+ which is a bit too restrictive.  At some point we should probably rework the whole contrib test system to add tests like this for the other contrib modules as well.

valentindusollier:
Yes, you're right. I think the right way to go for now is to remove it from `requirements.txt` and leave it as optional in `/testcontrib.py`. This way those who have the module installed can use it but `/testcontrib.py` works fine without `dotenv` as well. What do you think ?

uzlonewolf:
Yes, that sounds good.

valentindusollier:
Done ✅

jasonacox:
Thanks @valentindusollier !

PR #390, opened 2023-07-24
Fixed endless loop causing 100% cpu usage

michaelmittermair:
The sleep is outside the loop. This lead on a raspberry to hit 100% of cpu usage just for the gateway


mschlenstedt:
@michaelmittermair is absolutely right. I can confirm that.

jasonacox:
Thanks @michaelmittermair  !  🙏

PR #397, opened 2023-08-17
Various updates inspired by recent issues

uzlonewolf:
Most of these are fairly minor, though "Scan for new IP address then auto-IP is used and connection to last-known address fails" and "Make connection/key errors more descriptive" are pretty big.

### DIFF
--- a/tinytuya/core.py
+++ b/tinytuya/core.py
@@ -199,6 +199,7 @@ ERR_CLOUDRESP = 910
 ERR_CLOUDTOKEN = 911
 ERR_PARAMS = 912
 ERR_CLOUD = 913
+ERR_KEY_OR_VER = 914
 
 error_codes = {
     ERR_JSON: "Invalid JSON Response from Device",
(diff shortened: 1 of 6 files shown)


jasonacox:
Tests look clean. Updated docs and bump to v1.12.10.

jasonacox:
Thanks, @uzlonewolf ! 🙏

jasonacox:
https://pypi.org/project/tinytuya/1.12.10/

tinytuya/core.py
```diff
@@ -885,19 +929,23 @@ def _get_socket(self, renew):
                         "socket unable to connect (timeout) - retry %d/%d",
                         retries, self.socketRetryLimit
                     )
+                    err = ERR_OFFLINE
```
jasonacox, line 932: I don't know how I missed this. This stepped on the exception local variable `err` which it won't allow.  Regardless, we shouldn't have had `err` for the exception anyway since it is used within the function already.  Renaming to `e`.

See #402 - Thanks @simon-kuzin

PR #424, opened 2023-11-16
Require pyca/cryptography >= 3.1, and add tools/fake-v35-device.py

uzlonewolf:
An interesting use of fake-v35-device.py is setting `bcast_to` to the IP of either a SmartLife-running device or a gateway/hub device and changing `gwId`/`real_key` to that of one of your devices - doing this will cause said device to send v3.5 commands to this script, which you can then inspect and reverse engineer.  This is basically how I got v3.5 support working despite not having any v3.5 devices.

ttcorefunc.py is a "known good" copy of tinytuya/core.py (with most of the device stuff stripped out).  By forcing it to use PyCryptodome I was able to use fake-v35-device.py to test the pyca/cryptography functions in the real tinytuya/core.py.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,12 @@
 # RELEASE NOTES
 
+## v1.13.1 - Cryptography Version
+
+* PyPI 1.13.1
+* Require pyca/cryptography>=3.1 or fallback to PyCryptodome
+* Add `tools/fake-v35-device.py` script to tools
+* Allow pyca/cryptography to GCM decrypt without the tag (makes it match PyCryptodome) by @uzlonewolf in https://github.com/jasonacox/tinytuya/pull/424
+
 ## v1.13.0 - Crypto Library Update
 
 * PyPI 1.13.0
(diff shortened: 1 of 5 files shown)


jasonacox:
✅ Worked for my use case...

Was
```
DEBUG:TinyTuya [1.13.0]

DEBUG:Python 3.6.9 (default, Mar 10 2023, 16:46:00) 
[GCC 8.4.0] on linux
DEBUG:Using pyca/cryptography 2.1.4 for crypto, GCM is supported
```

This PR
```
DEBUG:TinyTuya [1.13.0]

DEBUG:Python 3.6.9 (default, Mar 10 2023, 16:46:00) 
[GCC 8.4.0] on linux
DEBUG:Using PyCryptodome 3.18.0 for crypto, GCM is supported
```

Testing others...

jasonacox:
Thanks @uzlonewolf !  I've pushed [v1.13.1](https://pypi.org/project/tinytuya/1.13.1/).

PR #445, opened 2024-01-17
add example for a string of XmCosy+ RGBW patio string lights

bikerglen:
I have some XMCosy+ RGBW patio string lights that permit the colors of each bulb to be set individually and independently of each other. This example takes a list of colors and assigns them to the bulbs in the string sequentially. The code encodes the list of colors into a string that can be sent to the bulbs using the set_value function with an index of 102. I've tested this with both a set of 15 patio string lights and a string of 6 flood lights. Both use Tuya controllers and are manufactured by the Shenzhen Bling Lighting Technology Company Limited. Index 102 is likely the mechanism used by the "DIY" mode's preview button in the Smart Life app.

### DIFF
--- a/examples/Contrib/XmCosyStringLight-example.py
+++ b/examples/Contrib/XmCosyStringLight-example.py
@@ -1,0 +1,113 @@
+# Works with XmCosy+ RGBW string lights.
+# Model # DMD-045-W3
+# FCC ID: 2AI5T-DMD-045-W3
+# Amazon SKU: B0B5D643VV
+#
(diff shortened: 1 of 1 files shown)


jasonacox:
Nice stand-alone example for the Tuya device Contrib repo. Thanks for the addition, @bikerglen !  🙏

bikerglen:
No problem. I just noticed I hard coded the number of colors in the chase to 6 in the for loop at the very end of main. I need to replace the 6 with a len(colors) and the 5-i with len(colors)-1-i on lines 110 and 111. I'll try to get a fix in tomorrow night after work.

PR #446, opened 2024-01-17
fix chase when the number of colors in the colors list is not exactly six

bikerglen:
I fixed (and tested) the xmcosy string light chase sequence code to work with numbers of colors besides just six.


jasonacox:
Thanks @bikerglen !

PR #451, opened 2024-01-29
Adding support for Presence Detector Device

mrioan:
This is the device: https://www.amazon.com/gp/product/B0BZCRNY14


jasonacox:
Hi @mrioan - Great addition and nice code.

Thanks for contributing to the project! 🙏

PR #468, opened 2024-02-24
Our purpose needs custom port + reduce delay when connection

unit-404:
When trying to use port attribute, we have huge delay due connection attempts. So, I propose to add port argument for direct connection to device by routers NAT's (custom port forwarding). 

(Sorry for my bad English).


jasonacox:
Thanks @unit-404 !  🙏  That's an interesting use case. I suspect not many will need it but it is a good enhancement. 

This will be included in v1.13.2.

PR #491, opened 2024-05-06
README:  Warn users about use of guest accounts

Hello1024:
Warn users against using guest accounts in the Smart Life App.

They can't be paired with Tuya's IOT platform, and attempting to pair them causes the guest account to be deleted and all paired devices forgotten, which can be a pain if you've already set everything up.


jasonacox:
Thanks for the doc update, @Hello1024 .Hopefully this will help others. 🙏

PR #502, opened 2024-05-26
Add CLI Support for pipx

### LINKED ISSUE #500: Allow to run with `pipx`
I don't know what it takes to make it work better with `pipx`, but it would be super nice if possible.

```console
$ docker run --rm -it ubuntu
$ apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends pipx
$ pipx install tinytuya
Note: Dependent package 'charset-normalizer' contains 1 apps
  - normalizer

No apps associated with package tinytuya. Try again with '--include-deps' to include apps of dependent packages, which are listed above. If you are attempting to
install a library, pipx should not be used. Consider using pip or a similar tool instead.
```

I am no Python expert. Maybe you just need to declare the tinytuya executable/binary in some manifest? Also, not sure about this "normalizer" thing.

jasonacox:
## v1.13.3

* Add support for `pipx install tinytuya` as raised by @felipecrs in https://github.com/jasonacox/tinytuya/issues/500 allowing for easier CLI use.
* Possible breaking change: Running `tinytuya` by itself will now produce a "Usage" page instead of running a scan.  Use `tinytuya scan` or `python -m tinytuya scan`.

Closes #500


uzlonewolf:
Here's the one I've been working on: https://github.com/uzlonewolf/tinytuya/tree/argparse

jasonacox:
We can focus on https://github.com/jasonacox/tinytuya/pull/503

Issue #507, opened 2024-06-11
Cannot connect to Smart Switch


measwel:
Good day,

The device shows up in the scan. The local key has been fetched with the wizard.

enlarger   Product ID = keyjup78v54myhan  [Valid Broadcast]:
    Address = 192.168.51.246   Device ID = bfadddf28e43521d4c7man (len:22)  Local Key = xxxxxxxxx  Version = 3.3  Type = default, MAC = 10:5a:17:dc:fd:32
    Status: {'1': True, '9': 0, '18': 0, '19': 0, '20': 2346, '21': 1, '22': 610, '23': 30696, '24': 17327, '25': 2410, '26': 0, '38': 'memory', '39': False, '40': 'relay', '41': False, '42': '', '43': '', '44': ''}

I read through all the issues concerning 901 connection errors, but not found a solution.

DEBUG:ERROR Network Error: Unable to Connect - 901 - payload: null
DEBUG:status() received data={'Error': 'Network Error: Unable to Connect', 'Err': '901', 'Payload': None}

I have 2 clues. 

1 : the core.py throws an error:

  File "...\AppData\Local\Programs\Python\Python312\Lib\site-packages\tinytuya\core.py", line 1009, in _get_socket
    if self.version >= 3.4:
       ^^^^^^^^^^^^^^^^^^^
TypeError: '>=' not supported between instances of 'str' and 'float'

2 : maybe other active connections are blocking access to the device?

I have a smart life app on my phone which shows the device. Disconnecting the device in the app does not work. When I disconnect without wiping data, the switch still loses its wifi settings. I have to repair it in the app to get it to show on the wifi network again. 

Just closing the smart life app on the phone also does not help. Could it be, because other home members also have that app on their phone and are possibly making a connection? If so, how should I solve that? Should I remove the device from all the smart life apps?

Thank you.

So I do not know how to make sure no other active connections are present.

uzlonewolf:
Hi @measwel,

Can you post your initialization code?  It looks like version is being set as a string (`"3.3"`) instead of a float (`3.3`) like it should.

You do not need to disconnect the device in the SmartLife app, simply closing the app is fine.  As for multiple connections, this is going to depend on what the device supports; some devices can handle 2-4 simultaneous connections while others only allow 1.

measwel:
I read device from device.json based on stored uuids of devices I am interested in.


def get_device_handle(device, type):
    device_id = device["id"]
    device_ip = device["ip"]
[...]

PR #511, opened 2024-06-23
Fix scanner force-scanning

uzlonewolf:
The move to argparse in the last release broke the force-scan IP list.


tinytuya/scanner.py
```diff
@@ -164,6 +164,7 @@ def __init__( self, ip, deviceinfo, options, debug ):
         self.timeo = 0
         self.resets = 0
         self.step = FSCAN_NOT_STARTED
+        self.try_v35 = False
```
jasonacox, line 167: Hey @uzlonewolf is this really only v35 or v34_and_v35? I'm trying to follow the logic/flow it seems like both, but I'm probably missing something.

tinytuya/scanner.py
```diff
@@ -334,7 +342,7 @@ def timeout( self, forced=False ):
                 print('ForceScannedDevice: Debug sock', self.ip, 'connect timed out!')
         elif self.step == FSCAN_INITIAL_CONNECT:
             if self.debug:
-                print('ForceScannedDevice: Debug sock', self.ip, 'socket send failed,', 'no data received' if forced else 'receive timed out')
+                print('ForceScannedDevice: Debug sock', self.ip, 'socket send failed,', 'no data received,' if forced else 'receive timed out,', 'current retry:', self.retries)
```
jasonacox, line 345: Good addition.

jasonacox:
Not sure if we want to address this because it is correct, but I suspect some (ok, even me) will specify a host address with netmask:

```
$ python3 -m tinytuya scan -f 10.0.1.1/24                                                                        pr/511

TinyTuya (Tuya device scanner) [1.14.1]

[Loaded devices.json - 43 devices]

Scanning on UDP ports 6666 and 6667 and 7000 for devices for 18 seconds...

    Option: Network force scanning requested.


    Running Scan...
ERROR: Unable to get network for '10.0.1.1/24', ignoring.
  File "/Users/jason/Code/tinytuya/tinytuya/scanner.py", line 931, in _generate_ip
    network = ipaddress.ip_network(netblock)
  File "/Users/jason/miniforge3/lib/python3.10/ipaddress.py", line 74, in ip_network
    return IPv4Network(address, strict)
  File "/Users/jason/miniforge3/lib/python3.10/ipaddress.py", line 1507, in __init__
    raise ValueError('%s has host bits set' % self)
ValueError: 10.0.1.1/24 has host bits set
```

tinytuya/scanner.py
uzlonewolf: It's only looked at when the device version is set to 3.4, so I guess `try_v35_with_v34` would be a more accurate name.

MarkCupitt:
@jasonacox @uzlonewolf 

Hi Guys, I grabbed server.py and scanner.py (I think it was only the two files) from @uzlonewolf's commits and ran them here, they did not seem to fix my issue, ie, did not see any v 3.5 devices, have attached log and devices file files for you, hope it helps

What I did notice last night I that 2 of the three devices that were not being seen by tinytuya eventually showed up on the webui, this morning, which was interesting. 

tinytuya scan -d >debug.log 2>&1
tinytuya scan -f 192.168.2.0/24 >force.log 2>&1

[debug.log](https://github.com/user-attachments/files/15949674/debug.log)
[devices.json](https://github.com/user-attachments/files/15949679/devices.json)
[force.log](https://github.com/user-attachments/files/15949671/force.log)

tinytuya/scanner.py
jasonacox: Got it! Yes, that makes sense. Thanks @uzlonewolf !

tinytuya/core.py
```diff
@@ -1654,6 +1654,7 @@ def add_dps_to_request(self, dp_indicies):
             self.dps_to_request.update({str(index): None for index in dp_indicies})
 
     def set_version(self, version): # pylint: disable=W0621
+        version = float(version)
```
jasonacox, line 1657: Good call! Yes, this surfaces as an issue every 3-6mo or so.

uzlonewolf:
@MarkCupitt that force.log is still using the old scanner.py, not this modified one, as shown by:

```
  File "/home/mark/.local/lib/python3.11/site-packages/tinytuya/scanner.py", line 902, in _generate_ip
    network = ipaddress.ip_network(netblock)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ipaddress.py", line 83, in ip_network
    raise ValueError(f'{address!r} does not appear to be an IPv4 or IPv6 network')
ValueError: ['192.168.2.0/24'] does not appear to be an IPv4 or IPv6 network
```
Replace `/home/mark/.local/lib/python3.11/site-packages/tinytuya/scanner.py` and try it again.

tinytuya/core.py
uzlonewolf: Yeah, I thought I'd sneak that one in there :)  I'm not marking it as 'closes #507' as I think there's more we can do to type-safe everything.

jasonacox:
@MarkCupitt if you want to pull the full repo down you could do this to test the PR:

```bash
# Clone
git clone https://github.com/jasonacox/tinytuya.git
cd tinytuya

# Grab PR
git pull origin pull/511/head

# Test
python3 -m tinytuya scan -f
```

jasonacox:
Try to stop the SmartLife app and see if the devices start broadcasting.

MarkCupitt:
@jasonacox @uzlonewolf Yep, I actually had disabled wifi for my phone with Smartlife on it, restarted all the device.

The pull request does address the 3.5 issue which was the root cause of my issue, and I accept that the devices are "different" and do something different with Broadcast packets.

Ill adjust my use case to work with this approach as I will know the IP addresses and can access them directly. Im comfortable with that, so I would say the pR is Ok

Thanks for the prompt fixes and responses, they are very much appreciated

MarkCupitt:
I'm going to call this issue successfully closed ..  as any remaining issues appear to be device-related. Thanks again ..

jasonacox:
I sure hope this isn't the start of a new device series/trend that don't broadcast discovery packets. 

Thanks for your help with this, @MarkCupitt !

uzlonewolf:
@jasonacox Do you want to add a "Start Force-scan" button to the server?

jasonacox:
> @jasonacox Do you want to add a "Start Force-scan" button to the server?

Oh, that's a great idea. Hum. Let me poke at that tonight. Might be worth including in this PR.

uzlonewolf:
I also fixed the "ValueError: 10.0.1.1/24 has host bits set" thing but it looks like I forgot to push it.  1 sec

tinytuya/scanner.py
```diff
@@ -928,7 +928,7 @@ def _generate_ip(networks, verbose, term):
         if tinytuya.IS_PY2 and type(netblock) == str:
             netblock = netblock.decode('latin1')
         try:
-            network = ipaddress.ip_network(netblock)
+            network = ipaddress.ip_network(netblock, strict=False)
```
jasonacox, line 931: So easy. ;-) thanks! 🙏

MarkCupitt:
@jasonacox as a suggestion, If you had a force scan update `snapshot.json`, perhaps add a key to the json like this:

```
poll_method: "broadcast"
or
poll_method: "poll"
```

with broadcast being the `default`, any other code that uses snapshot.json would know that the device was there, and that it needed to poll the device and not wait for a broadcast.

Secondly, adding a timestamp to the time it was last successfully accessed would allow for an last accessed age, thus allowing logic for retries on individual devices, should it pass a threshold, or then showing them as going off line

Lastly, if you were to add a new file, say "local_ips.json` something like this

```
[{
    "deviceid": "eb6cfead8a7ce1de12wvaf",
    "ip": "192.168.2.x"
  }]
```

You could build in a poll (on a timer)  like monitor.py into the scan only for non broadcast devices and be able to handle most edge cases I would think

Updating the UI to indicate if its a broadcast or poll and the age ( eg: formatted datatime -> 12 seconds ago) would also be very useful

Lastly, a wishlist request, UPdate the Offline devices from Devices.json so it was the same as the Online Table display 

This covers pretty much all the UI and oddball devices I can see form where I sit, and may make tinytuya more flexible 

I'm quite happy to contribute, but Python is not my first language, and its 15 years old, but feel free to ask

If you do agree to the above, I would be happy to write a separate mqtt server that uses the above (basically reads snapshot.json, from server.py) to publish values to mqtt on a regular basis, my feeling is that this is best left out of server.py, as its use case is very different, although I see there has been some thought on mqtt already, happy to contribute to that .. if you deem that a better approach

The end result would be a pretty cool setup where a process server.py was running in the background keeping snapshot.json up to date, then a second process that read snapshot.json, ( looked for file changes and reload would be nice) and polled the devices (threaded or asyncio) for status and emitted MQTT publish events

tinytuya/scanner.py
```diff
@@ -944,7 +944,9 @@ def _generate_ip(networks, verbose, term):
         for addr in ipaddress.IPv4Network(network):
             yield str(addr)
 
-def _print_device_info( result, note, term, extra_message=None ):
+def _print_device_info( result, note, term, extra_message=None, verbose=True ):
+    if not verbose:
+        return
```
jasonacox, line 949: I noticed while testing that `verbose=False` was not honored.

```python
scanner.devices(verbose=False, color=False, poll=False, forcescan="10.0.0.0/24",discover=False)
```

Printed all the devices.  I updated scanner to ignore prints if verbose=False.

jasonacox:
I added a "Force Scan" to the server which causes it to launch a `scanner.devices(forcescan=True, ...)` call.  I have the 3.5 Smartbulb on my network and force scan finds it, but not via discover (as we already found out). I will add the 3.5 broadcast packet (7000) to the server once we have it in the scanner.

tinytuya/scanner.py
```diff
@@ -1175,6 +1284,11 @@ def tuyaLookup(deviceid):
     if not scantime:
         scantime = 0 if ip_scan_running else tinytuya.SCANTIME
 
+    client_bcast_addrs = get_ip_to_broadcast()
+    for bcast in client_bcast_addrs:
+        addr = client_bcast_addrs[bcast]
+        client_ip_broadcast_list[addr] = { 'broadcast': bcast }
+
```
jasonacox, line 1291: Would it be simpler to remove this logic and have `send_discovery_request()` build the broadcast list itself (assume we always want it to broadcast to all interfaces)? Or do you think we will want to alter / prune the list it would use?

Either is fine... just selfishly wondering if I could just call `scanner.send_discovery_request()` without worrying about sending it any arguments. 😋

uzlonewolf: Yep, 2 reasons: so we only need to build the interface list/encrypt the payload/open the socket once, and so we can expand the list later if needed (i.e. we're using the single getmyIP() address and want to add any address we receive a broadcast on).

It shouldn't be hard to make the `send_discovery_request()` argument optional.

uzlonewolf:
Ok, I think that's everything on my list for now.

tinytuya/Cloud.py
```diff
@@ -396,8 +396,11 @@ def _get_all_devices( self, uid=None, device_ids=None ):
     def getdevices(self, verbose=False, oldlist=[], include_map=False):
         """
         Return dictionary of all devices.
-        If verbose is true, return full Tuya device
-        details.
+
+        Args: 
+            verbose - Returns raw JSON data from Tuya Cloud
+            oldlist - List of devices from previous run
+            include_map - Include the DPS mapping in the device list
```
jasonacox, line 403: I discovered I needed this better description of the args while adding the cloud sync button to the proxy. :-)

jasonacox:
LGTM - I'm going to do a few final tests and then merge and release v1.14.1.

jasonacox:
@uzlonewolf We may have a problem. I just tested on a system without netifaces or psutil and I'm getting an error for the basic scan.  After installing psutil, it works without any issue. We should add psutil to the setup requirements. I'll tests on a few other OS variants to make sure that works.

```
$ python3 -m tinytuya scan

TinyTuya (Tuya device scanner) [1.14.1]

[Loaded devices.json - 43 devices]

Scanning on UDP ports 6666 and 6667 and 7000 for devices for 18 seconds...

  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/__main__.py", line 124, in <module>
    scanner.scan( scantime=args.max_time, color=(not args.nocolor), forcescan=args.force, discover=(not args.no_broadcasts), assume_yes=args.yes )
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/scanner.py", line 1027, in scan
    devices(verbose=True, scantime=scantime, color=color, poll=True, forcescan=forcescan, discover=discover, assume_yes=assume_yes)
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/scanner.py", line 1636, in devices
    send_discovery_request( client_ip_broadcast_list )
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/scanner.py", line 218, in send_discovery_request
    iface['socket'].bind( (address,0) )
socket.gaierror: [Errno -5] No address associated with hostname


$ python3 -m tinytuya scan -f

TinyTuya (Tuya device scanner) [1.14.1]

[Loaded devices.json - 43 devices]

Scanning on UDP ports 6666 and 6667 and 7000 for devices for 18 seconds...

    Option: Network force scanning requested.

    NOTE: neither module netifaces nor module psutil are available, multi-interface machines will be limited.
           (Requires: `pip install netifaces` or `pip install psutil`)


    Running Scan...
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/__main__.py", line 124, in <module>
    scanner.scan( scantime=args.max_time, color=(not args.nocolor), forcescan=args.force, discover=(not args.no_broadcasts), assume_yes=args.yes )
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/scanner.py", line 1027, in scan
    devices(verbose=True, scantime=scantime, color=color, poll=True, forcescan=forcescan, discover=discover, assume_yes=assume_yes)
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/scanner.py", line 1636, in devices
    send_discovery_request( client_ip_broadcast_list )
  File "/home/jason/.local/lib/python3.10/site-packages/tinytuya/scanner.py", line 218, in send_discovery_request
    iface['socket'].bind( (address,0) )
socket.gaierror: [Errno -5] No address associated with hostname
```

jasonacox:
Trying:

pip3 install tinytuya==1.15.0.dev0

jasonacox:
So it seems that pstuil doesn't work well on Windows. I switched to netifaces.

pip3 install tinytuya==1.15.0.dev1

Successful tests on MacOS, Linux, RPi and Windows.

jasonacox:
Library: https://pypi.org/project/tinytuya/1.15.0

Server (in a can): jasonacox/tinytuya:1.15.0p13 - Run via docker, where HOST is the IP address of the host running the server (required to prevent force scan from scanning the /8 docker IP space :) 

```bash
docker run \
    -d \
    -p 8888:8888 \
    -p 6666:6666/udp \
    -p 6667:6667/udp \
    -p 7000:7000/udp \
    --network host \
    -e DEBUGMODE='no' \
    -e HOST='192.168.0.100' \
    -v $PWD/devices.json:/app/devices.json \
    -v $PWD/tinytuya.json:/app/tinytuya.json \
    --name tinytuya \
    --restart unless-stopped \
    jasonacox/tinytuya
```

jasonacox:
Nice! I'm wondering if we should reverse the dependency of netifaces. I tend to favor that to keep it clean, but having it install by default does give more functionality. Thoughts?

uzlonewolf:
I keep waffling on the netifaces dependency.  On one hand it's practically a requirement for multi-interface machines, but on the other how many of our users are actually using multi-interface machines?  I could go either way on it.

PR #517, opened 2024-07-08
Add netifaces as an install requirement

jasonacox:


jasonacox:
https://pypi.org/project/tinytuya/1.15.0/

PR #519, opened 2024-07-08
Fix scanner IP address binding when netifaces/psutil are not available

### LINKED ISSUE #516: Protocol 3.5 Device Discovery Updates
### Device discovery packets redux

Newer v3.5 devices do not send out unsolicited discovery broadcasts.  Instead, they listen for broadcasts from a client app and send their discovery packet directly to that client.

The client broadcast contains the payload `{"from":"app","ip":"192.168.1.42"}` and is sent to the subnet broadcast address (i.e. 192.168.1.255) on port 7000.  It is GCM encrypted the same way as broadcasts above.

When a device receives a client broadcast, it responds by sending a device discovery packet directly to the IP address specified in the client broadcast.  This again uses port 7000 and is GCM encrypted the same way as broadcasts above.  The device I used in my testing will send the device discovery packet out as a broadcast if the client app specifies "255.255.255.255" as the IP address, however I suspect this is an accident and is not something they intended to be used.

_Originally posted by @uzlonewolf in https://github.com/jasonacox/tinytuya/discussions/260#discussioncomment-9979656_

### LINKED ISSUE #522: relax requirement for netifaces
Recently a change was made to make netifaces a dependency. I've got some issues installing netifaces on my chromebook, and since I don't need the scanning part there, I'm ok if I don't have it.

I _think_ you've written the code in a way that means the package is still functional without netifaces, I can see:

```python
try:
    import netifaces # pylint: disable=E0401
    NETIFLIBS = True
except ImportError:
    NETIFLIBS = False
```

in `scanner.py` anyway. So I was wondering if you'd consider making `netifaces` and extra dependency. So users could install it without needing netifcases, something like:

```sh
pip install tinytuya # wouldn't have netifaces

pip install tinytuya[netiface] # if you need it.
```

Probably would have been nicer the othe way around, so users could opt out of netifiaces, while the default would be to include it, but I'm not sure pip has an opt-out approach, just an opt in one.

uzlonewolf:
Title.

Closes #516 
Also closes #522

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,7 @@ TinyTuya can also connect to the Tuya Cloud to poll status and issue commands to
 # Example Usage of TinyTuya
 import tinytuya
 
-d = tinytuya.OutletDevice('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', 'LOCAL_KEY_HERE')
-d.set_version(3.3)
+d = tinytuya.Device('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', 'LOCAL_KEY_HERE', version=3.3)
 data = d.status() 
 print('Device status: %r' % data)
 ```
@@ -276,8 +275,7 @@ import tinytuya
 """
 OUTLET Device
 """
-d = tinytuya.OutletDevice('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', 'LOCAL_KEY_HERE')
-d.set_version(3.3)
+d = tinytuya.Device('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', 'LOCAL_KEY_HERE', version=3.3)
 data = d.status()  
 
 # Show status and state of first controlled switch on device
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -23,8 +23,7 @@ TinyTuya can also connect to the Tuya Cloud to poll status and issue commands to
 # Example Usage of TinyTuya
 import tinytuya
 
-d = tinytuya.OutletDevice('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', 'LOCAL_KEY_HERE')
-d.set_version(3.3)
+d = tinytuya.Device('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', 'LOCAL_KEY_HERE', version=3.3)
```
jasonacox, line 26: Good with that too...

jasonacox:
Thanks @uzlonewolf !  

Tests looked good. Released as https://pypi.org/project/tinytuya/1.15.1/

PR #521, opened 2024-07-16
Update scanner.py

blitzu:
Here’s a modified version of the send_discovery_request function to include debugging and exception handling:

### DIFF
--- a/tinytuya/scanner.py
+++ b/tinytuya/scanner.py
@@ -199,7 +199,7 @@ def get_ip_to_broadcast():
     ip_to_broadcast['255.255.255.255'] = getmyIP()
     return ip_to_broadcast
 
-def send_discovery_request( iface_list=None ):
+def send_discovery_request(iface_list=None):
     close_sockets = False
 
     if not iface_list:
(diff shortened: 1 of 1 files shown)


tinytuya/scanner.py
```diff
@@ -208,32 +208,38 @@ def send_discovery_request( iface_list=None ):
...
-            iface['socket'].setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
-            iface['socket'].bind( (address,0) )
+            try:
+                print(f"Attempting to bind to address: {address}")
+                iface['socket'] = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)  # UDP
+                iface['socket'].setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
+                iface['socket'].bind((address, 0))
+            except socket.gaierror as e:
+                print(f"Failed to bind to {address}: {e}")
+                continue
```
jasonacox, line 223: Thanks for this PR @blitzu !  I'm wondering, why use `print()` instead of `log.debug()` and `log.error()` here?

blitzu: this is what chatgpt gave me.  i had an error on using scanner.py and this was the solution from chatgpt and it works

uzlonewolf:
I'm gonna vote "no" for this one.  PR #519 already fixes the underlying problem (instead of just hiding the error), print() should not be used here, and I like my spaces :rofl:

jasonacox:
> and I like my spaces 🤣

Ha! I didn't even see them until @blitzu's PR. We could use a style guide helper. I'm tempted to cherry pick just the spaces fix for this PR.  😝 

> PR https://github.com/jasonacox/tinytuya/pull/519 already fixes the underlying problem (instead of just hiding the error)

That's true.  I need to merge that and get into a minor rev release. 

@blitzu even though this PR will be addressed with #519 , we appreciate you raising this! Active community members and willing contributors like you are how we keep the project moving forward. Thank you!

uzlonewolf:
Personally I like having spaces inside brackets/parenthesis as I find it a lot more readable when scanning through code.

uzlonewolf:
Full fix in #519.

PR #525, opened 2024-07-30
Rewrite BulbDevice and rework set_multiple_values()

uzlonewolf:
This is a pretty big rewrite to BulbDevice.  The fact that it connects to the bulb and calls `status()` as soon as the version is set has always bugged me, so now it delays that until a command is sent.  If a user calls `status()` themselves before they send a command then it uses the result from that status call instead of calling it again.

Also included are modifications to core: a caching mechanism has been added so devices can try to use cached DP values instead of making calls to the device, and `set_multiple_values()` has been modified to try and detect devices which fall over if they get sent multiple DPs in a single request.  The caching only works when an active persistent connection is open; closing the connection clears the cache since we won't get notified when DPs change.

To access the cache, call `d.cached_status(nowait)`.  When `nowait=True` then the cached value is returned if it is available or `None` if it is not.  When `nowait=False` then the cached value is returned if it is available or a call to `d.status()` is made if it is not.

BulbDevice now leans on the cache to try and minimize the number of DPs set in `set_white()` and `set_colour()`.  This combined with the `set_multiple_values()` change is to help with bulbs that only accept 1 DP at a time, such as @Anonymouse83's in #504.

Todo: update documentation.

### DIFF
--- a/tinytuya/BulbDevice.py
+++ b/tinytuya/BulbDevice.py
@@ -78,17 +78,23 @@ class BulbDevice(Device):
         "24": "colour",
     }
 
-    # Set Default Bulb Types
-    bulb_type = "A"
-    has_brightness = False
-    has_colourtemp = False
-    has_colour = False
-
     def __init__(self, *args, **kwargs):
+        # Set Default Bulb Types
+        self.bulb_type = None
+        self.has_brightness = None
+        self.has_colourtemp = None
+        self.has_colour = None
+
         # set the default version to None so we do not immediately connect and call status()
         if 'version' not in kwargs or not kwargs['version']:
             kwargs['version'] = None
         super(BulbDevice, self).__init__(*args, **kwargs)
+
+    def status(self, nowait=False):
+        result = super(BulbDevice, self).status(nowait=nowait)
+        if result and (not self.bulb_type) and (self.DPS in result):
+            self.detect_bulb(result)
+        return result
 
     @staticmethod
     def _rgb_to_hexvalue(r, g, b, bulb="A"):
@@ -197,49 +203,39 @@ class BulbDevice(Device):
         
         return (h, s, v)
 
-    def set_version(self, version): # pylint: disable=W0621
-        """
-        Set the Tuya device version 3.1 or 3.3 for BulbDevice
-        Attempt to determine BulbDevice Type: A or B based on:
-            Type A has keys 1-5 (default)
-            Type B has keys 20-29
-            Type C is Feit type bulbs from costco
-        """
-        super(BulbDevice, self).set_version(version)
-
-        # Try to determine type of BulbDevice Type based on DPS indexes
-        status = self.cached_status()
-        if status is not None:
-            if "dps" in status:
-                if "1" not in status["dps"]:
-                    self.bulb_type = "B"
-                if self.DPS_INDEX_BRIGHTNESS[self.bulb_type] in status["dps"]:
-                    self.has_brightness = True
-                if self.DPS_INDEX_COLOURTEMP[self.bulb_type] in status["dps"]:
-                    self.has_colourtemp = True
-                if self.DPS_INDEX_COLOUR[self.bulb_type] in status["dps"]:
-                    self.has_colour = True
-            else:
-                self.bulb_type = "B"
-        else:
-            # response has no dps
-            self.bulb_type = "B"
-        log.debug("bulb type set to %s", self.bulb_type)
-
     def turn_on(self, switch=0, nowait=False):
         """Turn the device on"""
         if switch == 0:
+            if not self.bulb_type:
+                self.detect_bulb()
             switch = self.DPS_INDEX_ON[self.bulb_type]
         self.set_status(True, switch, nowait=nowait)
 
     def turn_off(self, switch=0, nowait=False):
         """Turn the device on"""
         if switch == 0:
+            if not self.bulb_type:
+                self.detect_bulb()
             switch = self.DPS_INDEX_ON[self.bulb_type]
         self.set_status(False, switch, nowait=nowait)
 
-    def set_bulb_type(self, type):
+    def set_bulb_type(self, type, has_brightness=None, has_colourtemp=None, has_colour=None):
         self.bulb_type = type
+
+        if has_brightness is not None:
+            self.has_brightness = has_brightness
+        elif self.has_brightness is None:
+            self.has_brightness = bool(self.DPS_INDEX_BRIGHTNESS[self.bulb_type])
+
+        if has_colourtemp is not None:
+            self.has_colourtemp = has_colourtemp
+        elif self.has_colourtemp is None:
+            self.has_colourtemp = bool(self.DPS_INDEX_COLOURTEMP[self.bulb_type])
+
+        if has_colour is not None:
+            self.has_colour = has_colour
+        elif self.has_colour is None:
+            self.has_colour = bool(self.DPS_INDEX_COLOUR[self.bulb_type])
 
     def set_mode(self, mode="white", nowait=False):
         """
@@ -249,6 +245,8 @@ class BulbDevice(Device):
             mode(string): white,colour,scene,music
             nowait(bool): True to send without waiting for response.
         """
+        if not self.bulb_type:
+            self.detect_bulb()
         return self.set_value( self.DPS_INDEX_MODE[self.bulb_type], mode, nowait=nowait )
 
     def set_scene(self, scene, nowait=False):
@@ -263,6 +261,9 @@ class BulbDevice(Device):
             return error_json(
                 ERR_RANGE, "set_scene: The value for scene needs to be between 1 and 4."
             )
+
+        if not self.bulb_type:
+            self.detect_bulb()
 
         if scene == 1:
             s = self.DPS_MODE_SCENE_1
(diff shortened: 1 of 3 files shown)


tinytuya/BulbDevice.py
```diff
@@ -99,18 +78,24 @@ class BulbDevice(Device):
...
         # set the default version to None so we do not immediately connect and call status()
         if 'version' not in kwargs or not kwargs['version']:
             kwargs['version'] = None
         super(BulbDevice, self).__init__(*args, **kwargs)
 
+    def status(self, nowait=False):
+        result = super(BulbDevice, self).status(nowait=nowait)
+        if result and (not self.bulb_type) and (self.DPS in result):
+            self.detect_bulb(result)
+        return result
```
jasonacox, line 97: I like this! clean.

tinytuya/BulbDevice.py
```diff
@@ -136,36 +121,33 @@ def _rgb_to_hexvalue(r, g, b, bulb="A"):
...
-            if len(hexvalue_hsv) == 7:
-                hexvalue = hexvalue + "0" + hexvalue_hsv
-            else:
-                hexvalue = hexvalue + "00" + hexvalue_hsv
+
+            # r:0-255,g:0-255,b:0-255|rgb|
+            for rgb_value in rgb:
+                value = int(rgb_value)
+                if value < 0 or value > 255:
+                    raise ValueError(f"Bulb type {bulb} must have RGB values 0-255.")
```
jasonacox, line 130: This is needed, but it occurs to me that we have a real mix of using python exceptions versus returning error payloads, or actually not doing any validation. This is more of a commentary on some of our global style/design we could address rather than specific to this PR. :)

tinytuya/BulbDevice.py
```diff
@@ -136,36 +121,33 @@ def _rgb_to_hexvalue(r, g, b, bulb="A"):
...
-            hexvalue_hsv = ""
-            for value in hsvarray:
-                temp = str(hex(int(value))).replace("0x", "")
-                if len(temp) == 1:
-                    temp = "0" + temp
-                hexvalue_hsv = hexvalue_hsv + temp
-            if len(hexvalue_hsv) == 7:
-                hexvalue = hexvalue + "0" + hexvalue_hsv
-            else:
-                hexvalue = hexvalue + "00" + hexvalue_hsv
```
jasonacox, line 157: I love that we are addressing this. This is original pytuya code and more complicated than it needs to be. Thank you.

tinytuya/BulbDevice.py
```diff
@@ -221,50 +203,40 @@ def _hexvalue_to_hsv(hexvalue, bulb="A"):
...
+        if has_colourtemp is not None:
+            self.has_colourtemp = has_colourtemp
+        elif self.has_colourtemp is None:
+            self.has_colourtemp = bool(self.DPS_INDEX_COLOURTEMP[self.bulb_type])
+
+        if has_colour is not None:
+            self.has_colour = has_colour
+        elif self.has_colour is None:
+            self.has_colour = bool(self.DPS_INDEX_COLOUR[self.bulb_type])
+
```
jasonacox, line 239: ❤️ this

tinytuya/BulbDevice.py
```diff
@@ -377,19 +362,8 @@ def set_hsv(self, h, s, v, nowait=False):
...
-            CONTROL,
-            {
-                self.DPS_INDEX_MODE[self.bulb_type]: self.DPS_MODE_COLOUR,
-                self.DPS_INDEX_COLOUR[self.bulb_type]: hexvalue,
-            },
-        )
-        data = self._send_receive(payload, getresponse=(not nowait))
-        return data
+
+        return self.set_colour( r * 255.0, g * 255.0, b * 255.0, nowait=nowait )
```
jasonacox, line 366: Also, clean API change. I think there is enough API change here we need to signal breaking change with the version number we pick.

tinytuya/BulbDevice.py
```diff
@@ -466,17 +446,29 @@ def set_white(self, brightness=-1, colourtemp=-1, nowait=False):
...
-            {
-                self.DPS_INDEX_MODE[self.bulb_type]: self.DPS_MODE_WHITE,
-                self.DPS_INDEX_BRIGHTNESS[self.bulb_type]: brightness,
-                self.DPS_INDEX_COLOURTEMP[self.bulb_type]: colourtemp,
-            },
-        )
+        payload = {
+            self.DPS_INDEX_BRIGHTNESS[self.bulb_type]: brightness,
+            self.DPS_INDEX_COLOURTEMP[self.bulb_type]: colourtemp,
+        }
```
jasonacox, line 452: Instead of `payload` should we move to `dps_values` or something like that to not confuse this with raw "payloads"?

tinytuya/BulbDevice.py
```diff
@@ -520,31 +519,35 @@ def set_brightness(self, brightness, nowait=False):
...
                 # for colour mode use hsv to increase brightness
                 if self.bulb_type == "A":
                     value = brightness / 255.0
                 else:
                     value = brightness / 1000.0
                 (h, s, v) = self.colour_hsv()
                 data = self.set_hsv(h, s, value, nowait=nowait)
+                msg += 'No repsonse from bulb.'
+            else:
+                msg += "Device mode is not 'white' or 'colour', cannot set brightness."
```
jasonacox, line 542: What about?...

```python
f"Unable to set brightness, device mode needs to be 'white' or 'colour' but reports {state["mode"]}"
```

tinytuya/BulbDevice.py
```diff
@@ -520,31 +519,35 @@ def set_brightness(self, brightness, nowait=False):
...
+        else:
+            msg += 'Unknown bulb state.'
 
         if data is not None or nowait is True:
             return data
         else:
-            return error_json(ERR_STATE, "set_brightness: Unknown bulb state.")
+            log.debug( msg )
+            return error_json(ERR_STATE, msg)
```
jasonacox, line 550: I mentioned how we use python exceptions in some cases and error responses in other place. But thinking about it, it seems we do this:

* If user provides invalid parameters (values) - raise exception
* If buld responds with error or invalid conditions - respond with error_json()

tinytuya/BulbDevice.py
```diff
@@ -588,33 +599,33 @@ def set_colourtemp(self, colourtemp, nowait=False):
...
         """Return colour as HSV value"""
-        hexvalue = self.status()[self.DPS][self.DPS_INDEX_COLOUR[self.bulb_type]]
+        if not self.bulb_type: self.detect_bulb()
+        hexvalue = self.cached_status()[self.DPS][self.DPS_INDEX_COLOUR[self.bulb_type]]
         return BulbDevice._hexvalue_to_hsv(hexvalue, self.bulb_type)
 
     def state(self):
         """Return state of Bulb"""
-        status = self.status()
+        status = self.cached_status()
```
jasonacox, line 628: I like the idea of caching. But what if the user is making changes to the device using the SmartLife app or other tool?

* We should definitely cache the device type and capabilities.
* Status or current state of device should be live (unless we provide a force=False)?

Thoughts?

tinytuya/core.py
```diff
@@ -820,7 +820,7 @@ def assign_dp_mappings( tuyadevices, mappings ):
 
 class XenonDevice(object):
     def __init__(
-            self, dev_id, address=None, local_key="", dev_type="default", connection_timeout=5, version=3.1, persist=False, cid=None, node_id=None, parent=None, connection_retry_limit=5, connection_retry_delay=5, port=TCPPORT # pylint: disable=W0621
+            self, dev_id, address=None, local_key="", dev_type="default", connection_timeout=5, version=3.1, persist=False, cid=None, node_id=None, parent=None, connection_retry_limit=5, connection_retry_delay=5, port=TCPPORT, max_simultaneous_dps=0 # pylint: disable=W0621
```
jasonacox, line 823: Should caching be an option (that can be turned off) in addition to persist=True?

Also, we should break up the super long line with some \n love.

tinytuya/BulbDevice.py
uzlonewolf: > But what if the user is making changes to the device using the SmartLife app or other tool?

Every device I've seen sends asynchronous updates when that happens, so as long as the user calls `.receive()` every now and then or sends a command with `nowait=False` then those changes will be picked up and cached.  This is also why caching is only enabled when `persist=True` and an active connection to the device is open.

tinytuya/core.py
uzlonewolf: I don't see a point in disabling it.  Cached data is only returned with `.cached_status()` so if you don't want cached data then just call `.status()`.  Enabling it without `persist=True` is going to cause the cached state to not match the device state if someone presses a button on the device or makes a change in the SmartLife app.

tinytuya/BulbDevice.py
uzlonewolf: I don't think there are any API changes?  Most of this is function de-duplication and code cleanup.  I'll double-check but I think the only user-visible changes are 1) manually reading `d.bulb_type` will read `None` until either `d.status()` is called or a command is sent  2) unknown bulbs now default to type 'B' instead of 'A' and  3) a `force` argument has been added to set_colour() and set_white()  to disable some of the new optimizations, however this feels like a dirty hack to me and I'm probably going to rework it to remove this.

jasonacox: Fair. Those are minor. Thinking of old code out there that may be impacted by those two things is very edge case. I'm good with a minor update. I had a stronger impression of the change when I was originally doing the review, but feel less concerned now.

tinytuya/BulbDevice.py
jasonacox: Sold.

jasonacox:
I think this is a good update @uzlonewolf - ready for non-draft PR IMHO.

uzlonewolf:
@jasonacox Do you happen to have an example status() response for a Type C bulb?  I want to check some of my logic for that bulb type before committing this.

jasonacox:
Hi @uzlonewolf , makes sense. However, I don't seem to have one. The Type C Bulb addition came from https://github.com/jasonacox/tinytuya/pull/54 by @davidlie (who may be able to test for us?) and I recall it was for the Feit Electric Smart Light Bulb. I see some on Amazon but not sure if that would be the C type. I thought I picked one up but scanned my devices and no Type C's. 😞

uzlonewolf:
Looking at that PR it looks to be a dimmer switch and not a bulb at all.  Using the Internet Archive I found some that look like they're probably it and ordered a couple new-old-stock off ebay.  We'll see in a few days.

jasonacox:
Yes, I can't recall why it said dimmer switch but PR was for a smart bulb. It wasn't that many years ago, but long enough fo rme to completely forget. 🤷 Thanks for doing the research. 🙏

uzlonewolf:
My dimmer switch came in Saturday and it does identify as a Type C bulb.  When it comes down to it, is there really a difference between a dimmer and a single-color (non-CCT) bulb?  Anyway, I also have some single-color as well as CCT bulbs arriving today.

The more I think about it the less happy I am with the entire "Type X" thing and am probably going to rewrite it to just use "has feature" (mode, brightness, colortemp, etc).  I may or may not try and use the DPS mapping (if it's available) instead of blindly guessing based off DP ID.

jasonacox:
Yes, I haven't been fond of the Type X thing either, but I don't know of a better way to represent both the DPS mapping as well as the range differences. 

https://github.com/jasonacox/tinytuya/blob/7d8922ee096b5519018fa068ffb698ef43a213cc/tinytuya/BulbDevice.py#L117-L127

uzlonewolf:
Last week I ended up rewriting BulbDevice and almost completely eliminated the Type X thing.  I left it in a few places for backwards compatibility and as a quick way to manually set the DP range, but it's otherwise gone.  The RGB/HSV range is now selected by a boolean, and the "has X capability" is set by whether a DP is defined or not.  I'll push an update once I have the DP detection finished.

I picked up an assortment of bulbs to test with.  The new Feit single-color white bulbs are v3.5 and act like what was formerly known as Type B but with a few DPs missing.  The older Feit dimmer uses DPs 1-3 with 3 being the "minimum dim" setting, while Geeni bi-color/CCT white only bulbs use DPs 1-3 with 3 being CCT.

obar:
Do your latest rewrite efforts (mentioned in the comment above) exist on github? I'd be interested in trying this out! Thanks so much to both of you for all of your work, it's both impressive and incredibly useful.

uzlonewolf:
Closing this and moving things over to #617

PR #558, opened 2024-11-12
Add IoT core service renewal info to the setup wizard in the README

lorumic:
The subscription to the `IoT core` service in the Tuya account expires, and after expiration the wizard won't be able to fetch data from the Tuya cloud anymore. I have added a paragraph to the guide in the README to clarify that it is possible to renew this subscription.


jasonacox:
Thanks @lorumic ! This is a helpful addition to our documentation. 🙏

PR #585, opened 2025-01-25
Fix WinError 10022: Handle invalid socket arguments in send_discovery_request

x011:
### **Overview**

This Pull Request addresses the `[WinError 10022] An invalid argument was supplied` error encountered when running the TinyTuya device scanner on Windows systems. The error occurs during the `sendto` operation in the `send_discovery_request` function within `scanner.py`.

### **Issue Details**

When executing the scan command, the following traceback is observed:

```
Traceback (most recent call last): File "<frozen runpy>", line 198, in _run_module_as_main ... OSError: [WinError 10022] An invalid argument was supplied
```

### DIFF
--- a/tinytuya/scanner.py
+++ b/tinytuya/scanner.py
@@ -234,8 +234,11 @@ def send_discovery_request( iface_list=None ):
 
         log.debug( 'Sending discovery broadcast from %r to %r on port %r', address, iface['broadcast'], iface['port'] )
         # the official app always sends it twice, so do the same
-        iface['socket'].sendto( iface['payload'], (iface['broadcast'], iface['port']) )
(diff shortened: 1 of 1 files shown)


jasonacox:
Makes sense. Thanks @x011 !

uzlonewolf:
I'm not a fan of just ignoring these errors like that.  Newer v3.5 devices require that broadcast to work or they'll never be found.

jasonacox:
Fair, the log error may not be strong enough to inform the user of a problem. I reverted for now.

@x011 When you saw this WinError, how did you address to make it work?

x011:
@jasonacox  I simply caught the error, and it worked after that. The scanner itself could find the devices, but due to this error, it would exit immediately.

PR #587, opened 2025-01-26
Add better error message when sending scanner broadcasts fails

uzlonewolf:
Modifies #585 to better warn users when sending broadcast discovery messages fails.  Because users may have VPNs and other non-broadcast-capable interfaces, the error is not displayed if sending on at least 1 interface succeeds.

### DIFF
--- a/tinytuya/scanner.py
+++ b/tinytuya/scanner.py
@@ -213,6 +213,8 @@ def send_discovery_request( iface_list=None ):
             addr = client_bcast_addrs[bcast]
             iface_list[addr] = { 'broadcast': bcast }
 
+    at_least_one_succeeded = False
+    bcast_error_messages = []
     for address in iface_list:
         iface = iface_list[address]
         if 'socket' not in iface:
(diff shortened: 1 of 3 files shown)


tinytuya/scanner.py
```diff
@@ -233,17 +235,23 @@ def send_discovery_request( iface_list=None ):
...
 
         if close_sockets:
             iface['socket'].close()
             del iface['socket']
 
+    if not at_least_one_succeeded:
+        if log.level != logging.DEBUG:
+            for line in bcast_error_messages:
+                log.error( line )
+        log.error( 'Sending broadcast discovery packet failed, certain v3.5 devices will not be found!' )
```
jasonacox, line 253: Clever. I like this approach.

jasonacox (approved):
Much better approach. It creates the right level of visibility into issues without stopping the scan.

jasonacox:
Version patch uploaded https://pypi.org/project/tinytuya/1.16.1/

PR #588, opened 2025-01-26
Ruff as linter, pyproject.toml for setuptools, uv.lock

tringenbach:
This PR does four things.

* Use Ruff instead of Pylint and configure it in pyproject.toml
* Configure setuptools with pyproject.toml
* Defines dependencies in pyproject.toml in a standards compliant manner (but with `uv` in mind)
* Adds a `uv.lock` file

Ideally a PR should only do one thing. I can break that out into multiple PRs or undo some of this if you want. 

My motivation for Ruff is speed. I kept making changes and having to wait a few seconds for the red squiggles in VSCode to disappear (or appear) with pylint.

I tried to keep the rules the same, but I wasn't entirely successful at that, so there's a few linter fixes in here too.

(Ruff can also act as a formatter, but I didn't enable or configure that.)

For setuptools, I don't recall why I started doing that. I think I was just reading about how build backends work and thought it would be nice to use the newer way of configuring it.

For uv, I was already using uv as pip replacement, and thought it would be cleaner to use it in project mode. Besides the lock file (which isn't strictly required), I didn't end up using any uv specific configuration, so it ought to work with poetry too, though I didn't test that.

I switched around how `version_tuple` and `__version__` were defined. That was because `setuptools` has to import the module, which requires all of it's dependencies to be installed at build time, if it's defined how it was. But if `__version__` is a literal, it doesn't need to do that, and my understanding is that is preferred. 

- Only tested locally, so the github workflow change is untested, and I also didn't test publishing, but the setuptools config changes might have accidentally messed with that
- I wasn't sure how to set up extras. I went with one extra per each cryptolib. There didn't seem to be a way to specify "at least one of these is required" though, so it's a little awkward
- In pyproject.toml, authors is a list, wasn't sure if anyone else should be added or not.
- If you use `uv`, you have to `uv sync --extra cryptography` or one of the other extras, if you want the project to actually run

I set the minimum Python version to 3.8, because that's what I saw in pylint.yml, but I see the others go down to 3.6, so perhaps I should change it to that?


jasonacox:
Thanks @tringenbach ! I'll take a look...

pyproject.toml
```diff
@@ -1,2 +1,161 @@
...
+]
+
+requires-python = ">= 3.8"
+
+[project.optional-dependencies]
+cryptography = ["cryptography>=3.1"]
+pycryptodome = ["pycryptodome"]
+pyaes = ["pyaes"]
+pycrypto = ["pycrypto"]
+
```
jasonacox, line 26: We had added the logic to setup.py for some small IoT devices/platforms that couldn't use some of these. I believe the intent was to allow the user to install one that did and the tinytuya iinstall would continue without insisting on cryptography. Yeah, I don't think it worked well. I do wonder how this would change that use case, if at all.

pyproject.toml
```diff
@@ -1,2 +1,161 @@
...
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+]
+
+dependencies = [
+    "colorama",
+    "requests",
+]
+
+requires-python = ">= 3.8"
```
jasonacox, line 19: I think this is fine. This aligns to end of life with python version (https://devguide.python.org/versions/)

@uzlonewolf  thoughts?

pyproject.toml
jasonacox, line 1: BTW, it was always on my list to get this done. Thank you!! 🙏 :)

uv.lock
jasonacox, line 1: Yes, uv makes sense. Not entirely sure how to  test this. :)

jasonacox:
> Use Ruff instead of Pylint and configure it in pyproject.toml

I understand the speed. I'll give it a good test.

jasonacox:
@tringenbach Can we break this apart. I would like to cherry pick the Ruff and UV support. A new PR may be easier? We may need to revisit the setuptools later. There are some compatibility issues with some platforms we need to work through before we commit to a switch.

jasonacox:
I appreciate the contribution @tringenbach but since there hasn't been a follow up, I'm closing this PR. Feel free to submit a fresh one or open an issue for discussion.

PR #608, opened 2025-04-22
Build test updates

uzlonewolf:
The 20.04 platform has been removed, so bump to 22.04.

Python 3.6 is not available on 22.04, so remove it.


jasonacox:
LGTM!  Thanks! 🙏

PR #609, opened 2025-04-27
Add cloud error handling and server recovery logic

jasonacox:
## v1.16.3 - Cloud Error Handling

* Add error handling in Cloud getdevices() function for edge case where old devices.json has corrupt or malformed device entries.
* Server p14 update: Add main loop logic to try to recover when exception occurs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,4 @@ mappings.json
 
 # test scripts
 server/r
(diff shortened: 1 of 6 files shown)


jasonacox:
https://pypi.org/project/tinytuya/1.16.3/

PR #612, opened 2025-04-28
Tool updates: pcap parse fix, new broadcast relay

uzlonewolf:
I've been having a bunch of WiFi performance issues lately, and I've been suspecting it has something to do with 60+ devices spewing broadcasts every 5 seconds.  If you didn't know, broadcasts are really bad for WiFi performance as they're all sent at the speed of the slowest client link on the whole network (which can be as low as 1 mbps).  Broadcasts can be disabled in UniFi by checking the "Block LAN to WLAN Multicast and Broadcast Data" box, however this brings up a problem: phone apps can no longer discover devices on the network.  To get around this, I wrote a simple broadcast-to-unicast relay.  Using unicast allows the packets to be sent at the full speed of the phone's connection (hundreds of mbps), and they are only sent when the app is actually in use.

To discover newer non-broadcasting v3.5 device you do need to add the MAC address of the computer running the relay to the "Excepted Devices" list, and to add new devices to the network the MAC of the app/phone must be added as well (I usually promptly remove it once I've finished adding devices).

I've only been running this for like a day and a half now, but I feel like blocking broadcasts has also helped with the "devices drop offline" problem as usually at least 1 device (out of the 60+) would have dropped offline by now.

In addition to the relay, this PR also has a couple updates to the pcap parser: one bug fix (variable got clobbered in `get_key()`,  an update so it can process relayed broadcasts, and an update so you don't have to provide the devices.json file if it's in the same or parent directory.

### DIFF
--- a/tinytuya/scanner.py
+++ b/tinytuya/scanner.py
@@ -252,6 +252,8 @@ def send_discovery_request( iface_list=None ):
                 log.error( line )
         log.error( 'Sending broadcast discovery packet failed, certain v3.5 devices will not be found!' )
 
+    return iface_list
+
 class KeyObj(object):
     def __init__( self, gwId, key ):
         self.gwId = gwId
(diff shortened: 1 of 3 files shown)


jasonacox:
Thanks @uzlonewolf !

PR #615, opened 2025-05-07
Fix initialization bug with python2 (embedded devices)

Ircama:
Thanks for this software.

This PR fixes a SyntaxError while importing *tinytuya* with Python2 (`import tinytuya`), e.g. for embedded devices not supporting Python3. Error:

```
Python 2.7.18 (default, May  7 2025, 19:04:48)
[GCC 13.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import tinytuya
  File "<stdin>", line 1, in <module>
  File "tinytuya/__init__.py", line 98, in <module>
    from .BulbDevice import BulbDevice
  File "tinytuya/BulbDevice.py", line 196
    raise ValueError(f"Unsupported bulb type {bulb} - unable to determine RGB values.")
                                                                                     ^
SyntaxError: invalid syntax
>>>
```

### DIFF
--- a/tinytuya/BulbDevice.py
+++ b/tinytuya/BulbDevice.py
@@ -193,7 +193,9 @@ class BulbDevice(Device):
             b = int(rgb[2] * 255)
         else:
             # Unsupported bulb type
-            raise ValueError(f"Unsupported bulb type {bulb} - unable to determine RGB values.")
+            raise ValueError(
+                "Unsupported bulb type {} - unable to determine RGB values.".format(bulb)
+            )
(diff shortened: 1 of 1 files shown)


jasonacox:
Hi @Ircama - thanks for this update. We no longer support Python 2.7 (see https://github.com/jasonacox/tinytuya/issues/376) but I don't see any reason to deny this. I'll tests.

jasonacox:
Thank you @Ircama ! 🙏

PR #617, opened 2025-05-09
Rewrite BulbDevice and rework set_multiple_values(), again

### LINKED ISSUE #432: Fast scrubbing through colours
I've built a similar GUI application to the Tuya Smart Life app with a hue colour wheel to quickly scrub through values.

Setting a single colour via `tinytuya` transitions immediately on the bulb, but when I scrub through many values quickly, it starts to lag and doesn't transition gracefully. The Smart Life app on the other hand can handle super-fast scrubbing, and the transitions are very smooth. Is the Smart Life app doing something different to achieve a more graceful effect?

In my code I'm opening a persistent connection, then sending `set_colour` for each colour change.

### LINKED ISSUE #549: Enhance Access to Color Functions
In the discussion [#548](https://github.com/jasonacox/tinytuya/discussions/548), it was suggested to promote the functions for hex to HSV and RGB conversions to officially supported methods. This will make it easier to access and use the color functions.

Relevant functions:
- `_hexvalue_to_hsv`
- `_hexvalue_to_rgb`
- `_rgb_to_hexvalue`

These functions are currently used as examples and should be officially supported to enhance usability.

### LINKED ISSUE #572: set dimmer level
I'm looking into this script to control my Tuya devices locally. I'm not familiar with Python, but so far it's looking promising and I'm already able to control my RGBWW lightbulbs (on/off/level) and outlets (smartplugs). Continuing with the code used for my lightbulbs I was also able to switch on/off my dimmers, but I'm not able to control the level. The following turns on the device, but does nothing with the brightness level:

```
d = tinytuya.BulbDevice('x', 'y', 'z', version=3.3)
d.set_brightness_percentage(50)
d.turn_on()
```

I'm familiar with the Tuya Cloud API and from this I've been using bright_value_v1 to set the level of my dimmers and bright_value_v2 to set the level of my lightbulbs.

Am I on the right track? Or should I take a completely different approach?

uzlonewolf:
I spent the last few days rewriting BulbDevice, then rewriting it again, and then again, and I now almost have something that I'm almost happy with :rofl:  With all the changes that have been added since #525 I decided to just start the PR over.  In addition to the comments in #525, my current issues are:

1) I feel that `set_white()`, `set_brightness()`, `set_colourtemp()`, `brightness()`, and `colourtemp()` should not exist, at least in their current form.  These functions accept/return different values based on what the bulb type is, forcing the user to know the implementation details for every bulb they use and makes them re-write their code should they ever swap out a bulb.  This goes against the entire point of abstracting away those details.  I would like to get rid of them and move everything over to their respective `*_percentage()` functions instead.

2) Non-underscore versions of `rgb_to_hexvalue()`, `hexvalue_to_rgb()`, and `hexvalue_to_hsv()` have been added, as well as a new `hsv_to_hexvalue()` for completeness.  Instead of a bulb type, these take a boolean `use_rgb` parameter.  This works fine for current bulbs, but my concern is what happens should a new format be introduced?  We can add an additional argument at that time, but I'm wondering if it would be better to replace that boolean with a selector that we can add options to later if needed (options would be, say, `rgb8` and `hsv16` for the current formats).  Thoughts?

3) I hate how  `brightness()`, `colourtemp()`, `colour_rgb()` and `colour_hsv()` all just assume a call to `status()` is going to be both successful and immediately returned (instead of being suck behind other responses that were queued because the user never calls `receive()`).  All the other functions return an error_json object if a call fails to return the expected result, but I don't see how we can do that here as the examples show stuff like `(r, g, b) = colour_rgb()` (which is going to blow up spectacularly and with a nearly-useless error message).  Should these functions be different and throw an exception instead, or should all functions be modified to throw exceptions on unexpected responses?

I don't think there are any breaking changes if people only used function calls, however if they got creative and tried to roll their own queries using/setting DPS_INDEX_* variables (*ahem* #432) then it's gonna break.  I removed all those variables as they're no longer needed, but can add them back if you think it will break less stuff.

The bulb capabilities detection should work pretty well with every bulb and dimmer I've seen, but I don't even want to think about what's going to happen if nowait=True is set before the detection finishes.

Speaking of nowait, I've been thinking about adding a non-blocking receive when `nowait=True` where all function calls will return a result if one is waiting, but will not wait for one.  This is to help with people who do stuff like #432 where they can have potentially hundreds of responses waiting in the receive buffer but are never going to call `receive()` to clear them.  Checking for waiting data without blocking can be done using the socket `peek()` function.  My only concern is users who have code which expects to read a response at some point but does not check for one when setting `nowait=True`.  Thoughts?

Beyond that, there is still some final cleanup needed before merging this, and lots of documentation needs updating too.

Closes #572 
Closes #549 
Closes #432


jasonacox:
Well crud, I may have stepped on this with https://github.com/jasonacox/tinytuya/pull/615.  I don't know why it couldn't resolve the merge.

uzlonewolf:
Yeah, I started to make a comment on that PR earlier but then figured it wouldn't be too hard to resolve the conflict.  This way Ircama gets some credit for contributing.

Looking into that build test failure now.

jasonacox:
Love this @uzlonewolf - I'm going to look through it in detail and test...

uzlonewolf:
Added a paragraph about music mode functions: I currently have 3 functions for music mode: set_musicmode(), unset_musicmode(), and set_music_colour(). I keep waffling on including set/unset; it basically just modifies (set) and reverts (unset) the .retry and .sendWait variables to try and make communications as fast as possible. In my testing the bulb has no problem with a 100ms update time, and sending the command and receiving the ACK averages around 55ms.

With regards to setting the bulb type/capabilities, there are basically 4 ways to set it:
1. Call `status()`, either explicitly (with `d.status()`) or implicitly via one of the functions which calls `state()` (which is, like, most of them)
2. Call `detect_bulb()` and either allow it to call `status()`, or provide a status response dps dict as its first parameter.
3. Call `set_bulb_type('x')` and pass it one of the old bulb types (A, B, or C).  It will assume the bulb has all possible capabilities for that bulb type and that they're using the default DP set.
4. Call `set_bulb_capabilities()` with a dict providing a custom DP mapping.  Only the `switch`, `brightness`, `value_min`, and `value_max` keys are required, all the rest are optional.

jasonacox:
Test are going well.  My list of tests were breaking comparisons with old code due to the additional debug print()'s so I moved those to log.debug() which I suspect was the intent.

I'll start working through your concerns...

>I don't think there are any breaking changes if people only used function calls, however if they got creative and tried to roll their own queries using/setting DPS_INDEX_* variables (ahem https://github.com/jasonacox/tinytuya/issues/432) then it's gonna break. I removed all those variables as they're no longer needed, but can add them back if you think it will break less stuff.

I'm big on "break less stuff" unless we have a compelling reason. I know I ended up using these myself and I suspect there are other integrations and derivatives that do as well. Since they don't hurt anything, let's put them back in.

jasonacox:
> I feel that set_white(), set_brightness(), set_colourtemp(), brightness(), and colourtemp() should not exist, at least in their current form. These functions accept/return different values based on what the bulb type is, forcing the user to know the implementation details for every bulb they use and makes them re-write their code should they ever swap out a bulb. This goes against the entire point of abstracting away those details. I would like to get rid of them and move everything over to their respective *_percentage() functions instead.

I'm keen to not break capability with existing integrations, however, I have an idea. I'll use `set_white()`. What if we moved to a polymorphic way to make arguments more abstract (bulb type agnostic)? My preference would be to use floating point (percentage) type to indicate an auto-map request but could be string, enum, etc.

```python
    def set_white(self, brightness=-1, colourtemp=-1, nowait=False):
        """
        Set white coloured theme of an rgb bulb.

       Args (automap):
            brightness(float): Value for the brightness (0.1 to 1.0) - automap to bulbtype range
            colourtemp(float): Value for the colour temperature (0.1 to 1.0) - automap to bulbtype range
        Args (legacy):
            brightness(int): Value for the brightness (A:25-255 or B:10-1000)
            colourtemp(int): Value for the colour temperature (A:0-255, B:0-1000).
            nowait(bool): True to send without waiting for response.

            Default: Max Brightness and Min Colourtemp
        """

        # determine if we need to convert from percentage
        if isinstance(brightness, float) or brightness == 0:
            if not 0.0 <= brightness <= 1.0:
                raise ValueError('set_white: The brightness needs to be between 0.0 and 1.0.')
           # convert to int based on bulb type
            value_max = self.dpset['value_max']
            value_min = self.dpset['value_min']
            brightness = int(value_min + (value_max - value_min) * brightness)
# ...


# Set the white colour with automap float values
bulb.set_white(0.5, 0.7)

# Set the white colour with legacy int values
bulb.set_white(128, 128)

```

Edge case is "1" in colourtemp (user would need to use 1.0) and may want to accomodate an int "0" for brightness.

jasonacox:
> Non-underscore versions of rgb_to_hexvalue(), hexvalue_to_rgb(), and hexvalue_to_hsv() have been added, as well as a new hsv_to_hexvalue() for completeness. Instead of a bulb type, these take a boolean use_rgb parameter. This works fine for current bulbs, but my concern is what happens should a new format be introduced? We can add an additional argument at that time, but I'm wondering if it would be better to replace that boolean with a selector that we can add options to later if needed (options would be, say, rgb8 and hsv16 for the current formats). Thoughts?

First, love this! Yes, we needed these as exposed functions. Also, I like where you are going. I think bool is too limiting. If we use something like `input_format` with options like rgb8 and hsv16 (or None to autoselect), that gives us flexibility.

jasonacox:
> I hate how brightness(), colourtemp(), colour_rgb() and colour_hsv() all just assume a call to status() is going to be both successful and immediately returned (instead of being suck behind other responses that were queued because the user never calls receive()). All the other functions return an error_json object if a call fails to return the expected result, but I don't see how we can do that here as the examples show stuff like (r, g, b) = colour_rgb() (which is going to blow up spectacularly and with a nearly-useless error message). Should these functions be different and throw an exception instead, or should all functions be modified to throw exceptions on unexpected responses?

Yes, I see what you mean. They were near-macro functions but we need a more robust way to support them. At a minimum, there needs to be some error logic in each of these (not just pass through) and I'm leaning toward exceptions for unexpected responses for the reasons you state.

jasonacox:
> Speaking of nowait, I've been thinking about adding a non-blocking receive when nowait=True where all function calls will return a result if one is waiting, but will not wait for one. This is to help with people who do stuff like https://github.com/jasonacox/tinytuya/issues/432 where they can have potentially hundreds of responses waiting in the receive buffer but are never going to call receive() to clear them. Checking for waiting data without blocking can be done using the socket peek() function. My only concern is users who have code which expects to read a response at some point but does not check for one when setting nowait=True. Thoughts?

It is a good idea but complex, potential breaking change. Maybe we add an argument `no_drain=False` for advanced use cases where user is wanting to handle the receive() themselves. I don't know a way to intuit that so it is still a breaking change. Hum... more thinking... ;) 

You know, if we added asyncio and offer up async tuned import classes, there is more ways we could handle this and wouldn't be limited with backward compatibility. But, oof... that would be some work.

uzlonewolf:
> due to the additional debug print()'s so I moved those to log.debug() which I suspect was the intent.

Those were actually supposed to be temporary and I had planned on deleting them completely, however I can see value in keeping them as debug's.  I'll reword them so it's obvious what they're showing.

uzlonewolf:
> I'll use `set_white()`. What if we moved to a polymorphic way to make arguments more abstract (bulb type agnostic)? My preference would be to use floating point (percentage) type to indicate an auto-map request but could be string, enum, etc.

I'm not really feeling it.  No existing code would use that and I feel for new code it would be better to just use the `set_*_percent()` version of those functions (which already support floats 0.0-100.0 (%)).  My current plan is to leave them but delete the references from the documentation with a code comment saying they are deprecated.

uzlonewolf:
> It is a good idea but complex, potential breaking change. Maybe we add an argument `no_drain=False` for advanced use cases where user is wanting to handle the receive() themselves. I don't know a way to intuit that so it is still a breaking change. Hum... more thinking... ;)

With how many functions would need to be modified, I don't think another argument in every call is the way to go.  A global switch such as `d.set_nowait_nonblock_recv(True)` (still working on that name...) to enable it would be better.

> You know, if we added asyncio and offer up async tuned import classes, there is more ways we could handle this and wouldn't be limited with backward compatibility. But, oof... that would be some work.

I have fundamental disagreements with how asyncio was implemented, so I'm not really feeling this.

I do, however, have a few ideas that I'm still fleshing out about how to make it completely non-blocking.  This is going to be a separate PR though :)

jasonacox:
> I'm not really feeling it. No existing code would use that and I feel for new code it would be better to just use the set_*_percent() version of those functions (which already support floats 0.0-100.0 (%)). My current plan is to leave them but delete the references from the documentation with a code comment saying they are deprecated.

Sold. I just don't want to break any existing integration. This would work.

jasonacox:
> d.set_nowait_nonblock_recv(True) (still working on that name...) 

That sort of describes it, but is not intuitive if you don't know the code. Probably ok since it is more advanced. But maybe `d.set_nowait_drain()` or `d.set_nowait_clear_buffer()` to indicate intention to drain the recv buffer even if you are not waiting?

And ...no love for asyncio? LOL 
> I do, however, have a few ideas that I'm still fleshing out about how to make it completely non-blocking. This is going to be a separate PR though :)

Oooh!!  I can't wait! ❤️

tinytuya/core/Device.py
```diff
@@ -103,11 +104,54 @@ def set_multiple_values(self, data, nowait=False):
...
+
         payload = self.generate_payload(CT.CONTROL, out)
-        return self._send_receive(payload, getresponse=(not nowait))
+        result = self._send_receive(payload, getresponse=(not nowait))
+
+        if result and 'Err' in result and len(out) > 1:
+            # sending failed! device might only be able to handle 1 DP at a time
+            for k in out:
+                res = self.set_value(k, out[k], nowait=nowait)
+                del out[k]
```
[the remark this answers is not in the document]
uzlonewolf: Section reworked.

uzlonewolf:
The hexvalue functions have been updated.  I started calling it `hex_format` but then changed it to `hexformat` to match the existing `hexvalue` variable name.  Options are "rgb8" and "hsv16" for `*_to_hexvalue()`, plus `None` for auto-detect in `hexvalue_to_*()`.

Debug logging messages updated.

The various functions which return current state values now raise a RuntimeError if the current state is unavailable for whatever reason.

tinytuya/BulbDevice.py
```diff
@@ -12,10 +12,10 @@
...
  Functions
     BulbDevice Class methods
-        rgb_to_hexvalue(r, g, b, use_rgb=False):
-        hsv_to_hexvalue(h, s, v, use_rgb=False):
-        hexvalue_to_rgb(hexvalue, use_rgb=None):
-        hexvalue_to_hsv(hexvalue, use_rgb=None):
+        rgb_to_hexvalue(r, g, b, hex_format):
+        hsv_to_hexvalue(h, s, v, hex_format):
+        hexvalue_to_rgb(hexvalue, hex_format=None):
+        hexvalue_to_hsv(hexvalue, hex_format=None):
```
jasonacox: Spotted one thing in docs: change to hexformat instead of hex_format

uzlonewolf: Yep, noticed right after committing.  Fixed.

uzlonewolf:
At this point I think all that's left is documentation and some commented-out code cleanup.

uzlonewolf:
I'm glad I reworked the examples because I had the scene stuff all messed up :sweat_smile:

Scenes fixed, and all 3 bulb examples re-written.

jasonacox:
I'll keep testing, but let me know when you think it is ready for merge. :)

jasonacox:
Combining makes sense as it is in support of the BulbDevice upgrades (even if it applies to more than that).

I'll start testing. For merge/release, we should bump version and we could stuff high level docs on changes (usage) in Release notes for now.

jasonacox:
Merged and released - View at:
https://pypi.org/project/tinytuya/1.17.0/

PR #637, opened 2025-08-05
Make colorama optional

### LINKED ISSUE #633: Removing python-colorama dependency
Hi
Still on my hunt to reduce the requirements on small devices (openwrt)

If we can't really remove cryptography, can we maybe remove or make optional such fancy things like "colorama" to save space for instance ?

uzlonewolf:
Closes #633 

Also fixes the newline in the Wizard region list that is in the wrong place.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,9 @@
 # RELEASE NOTES
 
+## v1.17.1 - Fix BulbDevice
+
+* Fix BulbDevice for non-bulb devices. This will allow BulbDevice to be used even for non-bulb devices, e.g. turn_on() & turn_off(), by @uzlonewolf in https://github.com/jasonacox/tinytuya/pull/620.
+
 ## v1.17.0 - BulbDevice Rewrite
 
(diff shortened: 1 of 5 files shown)


uzlonewolf:
Draft until I can test a few things and get the README updated.

grosjo:
Great ! 
Would that be in a next release (whl format) ?

uzlonewolf:
And with that, everything looks good from my end.

jasonacox:
Thanks @uzlonewolf, looks good!  I'll run the test and bump the version.

grosjo:
I get the error: 

```
# python pip-25.2-py3-none-any.whl/pip install --no-index tinytuya-1.17.3-py2.py3-none-any.whl 

WARNING: Disabling truststore since ssl support is missing
Processing ./tinytuya-1.17.3-py2.py3-none-any.whl
Requirement already satisfied: requests in /usr/lib/python3.11/site-packages (from tinytuya==1.17.3) (2.32.3)
INFO: pip is looking at multiple versions of tinytuya to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement colorama (from tinytuya) (from versions: none)
ERROR: No matching distribution found for colorama

```

PR #640, opened 2025-08-22
Add constructor option to specify configfile

blackw1ng:
Previously, CONFIGFILE was hardcoded to "tinytuya.json". This is now an option in the constructor to specify this, defaulting to the above.

This might be useful in settings where `tinytuya` is invoked in like cronjobs and we want to specify the location of the configfile ;)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,9 @@
 # RELEASE NOTES
 
+## 1.17.4
+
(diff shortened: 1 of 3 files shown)


jasonacox:
Thanks @blackw1ng !

Update released: https://pypi.org/project/tinytuya/1.17.4/

PR #645, opened 2025-09-06
v2.0.0 async

jasonacox:
## v2.0.0 - Async-First Architecture

This pull request introduces TinyTuya v2.0.0, a major release focused on a new async-first architecture that delivers improved performance, eliminates code duplication, and maintains full backward compatibility. The documentation has been updated to guide users through async features, and new usage examples and technical details are provided. There should be no breaking changes; all existing synchronous code continues to work unchanged.

**Async-First Architecture & Core Features:**

* All device implementation now lives in async classes (`DeviceAsync`, `OutletDeviceAsync`, etc.), with sync classes as thin wrappers for backward compatibility. This eliminates code duplication and streamlines future development. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R20-R41) [[2]](diffhunk://#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23cL3-R163)
* The new `AsyncRunner` utility ensures cross-platform compatibility, safe event loop management, and seamless delegation between sync and async APIs.
* Updates to the `README.md` detail async usage, concurrent device management, and provide example code for both sync and async workflows. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L124-R147) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R170-R238) [[3]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R458-R501)
* API documentation: lists new async methods, context manager support, and concurrent operation patterns for all major device types. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L192-R281) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L218-R327) [[3]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L238-R344)

**Tests:**

* New async example script `examples/monitor_async.py` demonstrates concurrent device monitoring and control using the async API.
* Release notes (`RELEASE.md`) provide a technical deep dive, migration guide, and highlight performance, testing, and compatibility assurances for v2.0.0.
* `pyproject.toml` is updated with pytest and coverage configuration to support improved testing and code quality.

**Documentation & Usability:**

* The updated documentation emphasizes the benefits of the async-first approach, provides migration guidance, and includes new sections for concurrent device management and async context managers. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L124-R147) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R170-R238) [[3]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R458-R501)
* References to new async examples are added throughout, helping users quickly adopt best practices. (F4525647L390R390)

**Backward Compatibility & Migration:**

* All changes are fully backward compatible; no migration is required for existing code, but users can gradually adopt async features for improved performance.

**Testing & Performance:**

* The codebase is validated with 59 comprehensive tests, real device validation, and performance benchmarking, ensuring robust async and sync operation. [[1]](diffhunk://#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23cL3-R163) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R3-R32)

Modified approach to #643 with credit to @uzlonewolf and @3735943886 

Claude Assisted 🚀 Code


tinytuya/core/__init__.py
```diff
@@ -16,3 +16,9 @@
 from .core import *
 from .core import __version__
 from .core import __author__
+
+# Conditionally import async modules, only available in Python 3.5 and above.
+import sys
+if sys.version_info >= (3, 5):
+    from .XenonDeviceAsync import *
+    from .DeviceAsync import *
```
jasonacox, line 24: Yeah, I know... basically anything less than 3.5 is going to NOP. I need to add an error and assert python 3.5+ support requirements.

3735943886:
You've done an excellent job getting so much done so quickly.
Great!

uzlonewolf:
That's a lot better, but I would still like to eliminate all functions in the non-async classes.  My idea was something like:

1) Create a class decorator somewhere in core:
```python
def make_class_synchronous(cls):
    print( 'patching class:', cls.__name__ )
    for func in dir(cls):
        atr = getattr(cls, func)
        if (not func.startswith("__")) and callable(atr) and (func != 'call_coro_synchronously'):
            print('patching function:', func, type(atr))
            # emulate a function @decorator
            setattr(cls, func, cls.call_coro_synchronously(atr))
    return cls
```

2) Create a function decorator in XenonDevice to run the coroutine synchronously:
```python
    @classmethod
    def call_coro_synchronously(cls, func):
        def wrapper(*args, **kwargs):
            result = func(*args, **kwargs)

            # if the function is async it will return a coroutine object
            if type(result).__name__ ==	'coroutine':
                # await it here                                                                                                                                                                         
                try:
                    loop = asyncio.get_running_loop()
                except RuntimeError:
                    return asyncio.run(result)
                else:
                    return loop.run_until_complete(result)

            # else it was not an async function so just return the result
            return result
        return wrapper
```

3) Lastly, replace all the non-async classes with:
```python
@make_class_synchronous
class Device(DeviceAsync):
    pass

@make_class_synchronous
class BulbDevice(BulbDeviceAsync):
    pass

@make_class_synchronous
class OutletDevice(OutletDeviceAsync):
    pass

etc...
```
That's it!

With my "Step 2" I was planning on keeping it simple like above, but there's no reason we couldn't instantiate your AsyncRunner there instead.  I.e.:
```python
class XenonDevice(object):
    #put it here so there's only 1 instance class-wide
    _async_runner = AsyncRunner()

    @classmethod
    def call_coro_synchronously(cls, func):
        def wrapper(*args, **kwargs):
            result = func(*args, **kwargs)
            if type(result).__name__ ==	'coroutine':
                # await it here
                return cls._async_runner.run(result)
            return result
        return wrapper

    def __init__(
...
```

uzlonewolf:
And of course it doesn't work like I expected.  Hmmm...

uzlonewolf:
While trying to figure out why my code wasn't working I discovered BulbDeviceAsync needs a _ton_ more await's mindlessly sprinkled throughout 👀🤣.  Basically, if it has a `nowait` argument it needs an `await`.

uzlonewolf:
Ok, I had a look, and I now have no idea what anything is doing anywhere.  WTF is `XenonDeviceAsync._initialize_payload_dict()`, why does it only contain device22+zigbee, why isn't it using the global `payload_dict`, and why is it making a `.copy()` of a dict it just created?

The previous issue I ran into in my bulb test is because both Auto-IP and key-from-devices.json are handled by `initialize()`, which is called from `@classmethod create()` and/or `__aenter__()`, and I have no idea where/how either of those are called.

jasonacox:
Oh no!  Ha! 😆 You're absolutely right @uzlonewolf. I had it on my list to test AutoIP and never did. 😞  Basically each device class needs to call `self._runner.run(self._async_impl.initialize())` - I'm not sure if that is the best approach to wrap it, but it seems to work. Got my bulb music working, but I don't have a 3.5 so also a problem with my test platform.

tinytuya/core/XenonDevice.py
```diff
@@ -33,189 +29,28 @@ def find_device(dev_id=None, address=None):
...
+    return runner.run(find_device_async(dev_id, address))
 
-    if ret is None:
-        ret = {'ip':None, 'version':None, 'id':None, 'product_id':None, 'data':{}}
-    log.debug( 'find() is returning: %r', ret )
-    return ret
+def device_info(dev_id):
+    """Get device info from devicefile"""
+    runner = AsyncRunner()
+    return runner.run(device_info_async(dev_id))
```
3735943886, line 38: It looks like `find_device` and `device_info` are already defined as synchronous methods in `XenonDeviceAsync`. Is it necessary to wrap the synchronous versions in an `async` function using `to_thread`, and then wrap that again with an `AsyncRunner`? It seems simpler to just import and use the existing synchronous versions directly.

jasonacox: I like it. Your suggestion would be:

```python
# In XenonDevice.py

from .XenonDeviceAsync import find_device as _sync_find_device, device_info as _sync_device_info

def find_device(dev_id=None, address=None):
    """Scans network for Tuya devices with either ID = dev_id or IP = address"""
    return _sync_find_device(dev_id, address)

def device_info(dev_id):
    """Get device info from devicefile"""
    return _sync_device_info(dev_id)
```

I had a discussion with Claude about it as well. He agreed the simplification would work, but still liked the existing method more. However, his arguments aren't very strong: "Since we are going async-first architecture and if we made those functions async in future, would need to refactor." 

He did admit that your approach was also cleaner and slight more efficient. 😛

jasonacox:
Running regression/simple_async_test.py, I noticing that while the results were incredibly fast (subsecond for 30 devices), the results were pretty terrible (30% devices with bad payloads). The `_send_receive()` was not actually doing any retries. I added code to address that (mirrors some of what we had before). There may be a better way but after the change, it is solid, if not even a little better than the current library (was getting some failures on old lib for some of my finicky devices, but 100% success on new).

I want to look more at the stream handlers. Ideally once you connect to a device, we could have a buffer which maintains a digital twin state, via messages (udpates) and sends heartbeats to keep it alive for the users. Responses get paired with the commands sent. Anyway, an idea I want to explore.

I'm offline for a bit. Feel free to test, review, etc.

3735943886:
I've added additional commits in #646. The current async wrapper was creating and destroying event loops on every sync call, which caused persistent operations to fail when the `persist=True` argument was used.

uzlonewolf:
Now for some random ramblings.  Coming from a C and PHP background, I've always used socket `select()` to allow connecting to multiple devices at once.  My multi-server has been using it to comfortably talk to 57 devices for many months now, all in a single thread.

asyncio, on the other hand, has always felt like the cheap crap you get when you can't be bothered to spend the effort to do it right.  It has all the pitfalls of multi-threaded programming (variables can randomly change on you between 2 lines of code) with none of the advantages (Python's Global Interpreter Lock [GIL] means you will never have more than 1 thread running at any time).

That said, my multi-server has some issues (especially around making the initial connection to devices) and needs some rewriting to handle v3.5 devices, so I've been thinking about rewriting it to work with asyncio instead.

I had some ideas about making it so you can give devices a callback for received data and then forget about it, and it will automatically do housekeeping (such as sending keep-alives) for you.  The more I thought about it though the more  I realized asyncio will not solve the big problem I was running into with my `select()` version - some Device classes send commands and need the response to that exact command (and only that command), and may have a sequence of commands/responses that cannot have additional commands slipped in between them by a different thread.  Though now that I'm writing this out I realize that can be easily solved by having a device lock that critical commands can acquire...  Hmm, I think the idea of converting my multi-server to asyncio is growing on me.

3735943886:
I created and tested some code for this, and it seems to work correctly. However, I wasn’t able to use the `__getattribute__`, since the `AsyncWrapper` wraps methods and makes it difficult to obtain the correct name. For now, I’ve only implemented `status`.
And because I don’t have a deep understanding of Tuya communications, I’m not sure whether calling `receive()` immediately after sending a packet produces the same result as using `nowait=False`, or whether it requires protection against race conditions—which I don’t yet know how to handle.

`Device.py`
```python
# ...

class Device(XenonDevice):
    # ...

    def __init__(self, *args, **kwargs):
        # ...

        # Newly added section
        self.register_data_handler(None)
        self._runner.run(self._async_impl._callback_receive_loop(), nowait=True)

    # Newly added method
    def receive(self):
        return self._runner.run(self._async_impl._callback_receive_internal())

    # Newly added method
    def status(self, nowait=False):
        self._runner.run(self._async_impl.status(nowait=True), nowait=True)
        if not nowait:
            return self.receive()
```

`DeviceAsync.py`
```python
    # Newly added method
    def register_data_handler(self, data_handler):
        self._callback_data_handler = data_handler

    # Newly added method
    async def _callback_receive_internal(self):
        async with self._callback_receive_condition:
            await self._callback_receive_condition.wait()
            return self._callback_receive_packet

    # Newly added method
    async def _callback_receive_loop(self):
        """
        Internal method to process sync requests in the background.

        This method handles incoming messages by enqueuing them for a synchronous process.
        It runs in the background to ensure the main thread remains responsive.

        Args: None
        """
        self._callback_receive_packet = None
        self._callback_receive_condition = asyncio.Condition()
        while True:

            _received = None
            try:
                _received = await self.receive()

                async with self._callback_receive_condition:
                    self._callback_receive_packet = _received
                    self._callback_receive_condition.notify_all()

                if self._callback_data_handler:
                    loop = asyncio.get_event_loop()
                    asyncio.ensure_future(loop.run_in_executor(None, self._callback_data_handler, _received))

            except:
                # Exception handling here
                pass
```

`test.py`
```python
import tinytuya

d = tinytuya.Device(
    dev_id='eb7ac8a28ceab00dfacwdk',
    address='10.10.240.39',
    local_key='1SmG@L9VIUS70?j(',
    version=3.4,
    persist=True)

def h(pckt):
    print(f"handler: {pckt}\n")

d.register_data_handler(h)

time.sleep(1)
print(f"status(nowait=True): {d.status(nowait=True)}\n")

time.sleep(1)
print(f"received(): {d.receive()}\n")

time.sleep(1)
print(f"status(): {d.status(nowait=False)}\n")

time.sleep(1)
```

Results
```bash
$ python3 test.py
status(nowait=True): None

handler: {'dps': {'1': False, '2': False, '9': 0, '10': 0, '38': 'memory', '40': 'relay', '41': False, '42': '', '43': '', '44': ''}}

received(): None

handler: None

status(): {'dps': {'1': False, '2': False, '9': 0, '10': 0, '38': 'memory', '40': 'relay', '41': False, '42': '', '43': '', '44': ''}}

handler: {'dps': {'1': False, '2': False, '9': 0, '10': 0, '38': 'memory', '40': 'relay', '41': False, '42': '', '43': '', '44': ''}}

```

jasonacox:
Sold! 🤣 

* Legacy Stability - I love the idea of keeping the sync aspects of the library 100% intact as they are today for all the reasons you both say. We so rarely change that code at this point, I don't see the double maintenance of code being a big burden. To @3735943886 point, we have over 200k downloads of the library PER MONTH so there is a sizeable community out there that is very dependant up on this module, with 282 projects directly connected to our efforts here. There is a responsibility to keep it stable.
* Asyncio Native - I also LOVE the idea of starting **new** on the async classes using asyncio and not try to honor the old API but instead lean in to the power of async (@uzlonewolf 's examples of better message handling). As I was building the wrappers trying to minimize redundant code, I kept thinking this was the wrong approach even if it we ever got to 1:1 API preservation.

I'm going to suggest a soul sucking provocative challenge. Success on getting true async classes and asyncio functions in place would mean that scanner.py can be rewritten to use that class to accomplish the same performance. It has always bugged me that scanner.py is essentially a completely separate algorithm from the library and how most people would use the library. Don't get me wrong, @uzlonewolf, did a brilliant job and it still impresses me on how well it performs. But ideally the code could be pushed into the class for anyone to use in a similar away. I think asyncio can help us get there.

Action 

* Create a legacy branch - https://github.com/jasonacox/tinytuya/tree/v1-Legacy - Regardless of approach, it was my intention to preserve a v1 legacy branch and releases (bugfix / security fix only).
* Create a new v2 branch and PR (https://github.com/jasonacox/tinytuya/tree/v2-async) - I'm happy to abandon this PR and v2.0.0-async branch in favor of this new approach. 

Thoughts? If you both agree, I'll add @3735943886 as a collaborator to be able to work in the same PR.

jasonacox:
This was fun! Now for the real story... 😀

PR #646, opened 2025-09-09
v2.0.0 async: fix

3735943886:
This PR reverts the `XenonDeviceAsync` class to a base state, mirroring the `XenonDevice` class with minimal socket communication patches to establish a robust foundation for upcoming changes, **though the revised socket code may still have undiscovered bugs, which will be fixed as they are discovered.**


3735943886:
Refactor: Use single event loop thread for persistent connections

The previous `AsyncRunner` used a `ThreadPoolExecutor` to run async tasks. This created a new event loop for each call, making it impossible to maintain stateful connections (i.e persist=True) across multiple calls.

This commit refactors the `AsyncRunner` to use a single, dedicated background thread that runs a persistent event loop. Async tasks are now submitted to this long-running loop.

This change enables stateful operations and improves performance for frequent calls by eliminating loop setup/teardown overhead.

PR #649, opened 2025-09-14
v2.0.0 with Async Architecture

jasonacox:
## v2.0.0 - Async Architecture Introduction (MAJOR REVISION)

This major release introduces the foundation for native asyncio-based device communication while fully preserving the existing synchronous API for backward compatibility.

Highlights:
* Version bump to 2.x to signal new async subsystem (legacy sync classes unchanged).
* Planning document `ASYNC.md` added (vision, goals, milestones for XenonDeviceAsync & related classes).
* No behavioral changes to existing synchronous code paths in this initial 2.0.0 tag.
* Future minor releases (2.1.x+) will add new async classes and examples without removing sync support.

Compatibility:
* Existing imports and synchronous usage continue to work (API surface of 1.x retained).
* New async classes will live alongside current modules (no name collisions) and require explicit opt‑in.
* Officially removed Python 2.7 support.

Migration Guidance:
* You can adopt async incrementally—no action required if you stay with sync API.
* When async classes land, prefer `await device.status_async()` patterns in event loops for concurrency gains.

See [ASYNC.md](https://github.com/jasonacox/tinytuya/blob/dbff4723c51834e8754c4c9e1291942ba89b64bf/ASYNC.md) for roadmap details.

Re: #645 #646


3735943886:
That looks like a very long way to go. 😓

The #645 v.2.0.0-async branch works well for me with my small patches from #646. I've been running it in my real environment for over 8000 minutes with around 100 devices connected. There have been no issues so far with resource leaks, deadlocks, or delays—just an instant response without threading, and so on.

jasonacox:
Yes, but I think it can be fast. And yes, @3735943886 - absolutely! Your code would be a good place to start. 

In the #646 branch have you been using the direct Async classes or the wrappers?

Would you be willing to move those *Async classes over to this branch?

3735943886:
As a first step, I've committed **`XenonDeviceAsync`** as a foundational class for upcoming **`v2-async`** support, with minimal socket patches. It seems quite stable so far, likely because the original **`XenonDevice`** class was so robust. My only concern is whether my revised socket communication code works without issue.

> In the #646 branch have you been using the direct Async classes or the wrappers?
>
> Would you be willing to move those *Async classes over to this branch?

I'm currently running a **`DeviceAsync`**-based approach for 24/7 asynchronous multi-device monitoring. You can view the code I'm using at this link: [https://github.com/3735943886/tuya2mqtt/blob/async-beta/core/tuya2mqtt.py](https://github.com/3735943886/tuya2mqtt/blob/async-beta/core/tuya2mqtt.py)

I've also tested some synchronous **`v1`** examples and shared my feedback in the comments on **PR #646**.

jasonacox:
> As a first step, I've committed XenonDeviceAsync as a foundational class for upcoming v2-async support, with minimal socket patches

Perfect! Are you able to push to [v2-async](https://github.com/jasonacox/tinytuya/tree/v2-async)?

I may have misunderstood the discussion in #645 but I came away thinking (and agreeing) that we want to keep the existing sync code in tact. The attempt now will be to create an extension to the library that is async. It looks like you closed and deleted #646 - but likely we could start with some of your fixes there. My proposal is to start with copying in:

Classes: 
* XenonDeviceAsync
* DeviceAsync

But I'm also wondering, do we need both? That inheritance paradigm came from pytuya and I often debated on why and where it made sense to put new functions.

Perhaps we just use XenonDeviceAsync and rename it as DeviceAsync and fold in the "Device" functions into that same class. Then we derive Outlet, Bulb, Cover, etc., from there.

```
+------------------------------+            +---------------------------+
|  DeviceAsync (base)          |            | MessageHelper (shared)    |
|  - state machine             |<--calls--> |  pack/unpack (sync funcs) |
|  - connection supervisor     |            |  crypto helpers           |
|  - protocol v3.1..v3.5       |            +---------------------------+
|  - send queue (asyncio.Queue)|
|  - recv task (reader loop)   |            +---------------------------+
|  - handshake coroutine       |<--uses---->| Crypto (AESCipher)        |
+--------------+---------------+            +---------------------------+
               | derives
    +----------+-----------+
    | Async Device Mixins  |
    | (Outlet/Bulb/etc.)   |
    +----------------------+
```

No sacred cows here. Love it, hate it? Cast your ideas out. 😁

uzlonewolf:
> I may have misunderstood the discussion in #645 but I came away thinking (and agreeing) that we want to keep the existing sync code in tact.

I wanted to temporarily keep it untouched for the first couple of releases until the async version has proven itself in the real-world, and then replace it with a wrapper later.

> Perhaps we just use XenonDeviceAsync and rename it as DeviceAsync and fold in the "Device" functions into that same class. Then we derive Outlet, Bulb, Cover, etc., from there.

Sounds good to me.  I think everything already derives from Device anyway, I don't remember anything using XenonDevice directly.

> Success on getting true async classes and asyncio functions in place would mean that scanner.py can be rewritten to use that class to accomplish the same performance. It has always bugged me that scanner.py is essentially a completely separate algorithm from the library and how most people would use the library.

Yes, I planned on rewriting the scanner to use the new API once said API has stabilized.  This is what, the 4th async PR?  I didn't want to rewrite it while things were still changing so much.  That said, the force-scan requires brute forcing both the device version and the local key, so it is always going to be a separate algorithm compared to code which already knows both.

uzlonewolf:
Ok, I think I have my callback idea mostly fleshed out.  Please note that this does not prevent you from sending/receiving as you've always done, this callback method is an option in addition to that.

```python
import asyncio
import tunytuya

async def data_handler( device, data, tuya_message ):
    # do something with the received data here
    print( 'in data handler', device.id, data )
    print( 'device list', devices )

async def cmd_handler( device, tuya_message ):
    # received a payload-less TuyaMessage, you can check the retcode here if you want
    pass

async def connected( device, error ):
    # connection to device established, or failed
    if not error:
        await device.status()

devices = {}

async def main():
    async with asyncio.TaskGroup() as tg:
        # start up a background scan job in case we have multiple devices using auto-ip
        scanner = tinytuya.scanner.background_scan()
        scanner_task = tg.create_task( scanner.run_task() )

        for devdata in device_list:
            d = tinytuya.DeviceAsync( devdata )

            d.register_data_handler( data_handler )
            d.register_command_handler( cmd_handler )
            d.register_connect_handler( connected )
            d.register_scanner( scanner )

            task = tg.create_task( d.run_task() )

            devices[d.id] = (d, task)

    # The await is implicit when the context manager exits.

asyncio.run( main() )
```

Of note is the new scanner interface.  The problem with the current scanner is it does not handle multiple devices with auto-ip well.  You cannot simply fire up multiple instances (i.e. 1 for each device) because only 1 UDP receiver can receive broadcast packets at a time.  Kicking off a single thread and then having each device register the ID of interest fixes that.

uzlonewolf:
> I believe providing a separate `@classmethod create(...)` is a reasonable design choice for users who want to initialize the device without using the `async with` syntax.

But where/who is supposed to call it?  I am not using `async with` in the callback example I posted above and ended up needing to call `initialize()` myself since `create()` was never called.

3735943886:
The `@classmethod create(...)` method is intended to be called by end users as an alternative to the `async with` context manager for initialization.

The previous `README` in PR #645 documented it, but it's not available in the current branch. We'd eventually need restoration the detailed documentation.

```python
# Recommended: Using async context manager
async with tinytuya.DeviceAsync(...) as device:
    # Use the device inside this block
    # Resources are automatically cleaned up when the block is exited.

# Alternative: Using the factory method
device = await tinytuya.DeviceAsync.create(...)
try:
    # Use the device
finally:
    await device.close()  # You must manually call close() to release resources.
```

uzlonewolf:
In this particular case I think the old behavior was actually a bug - why should a device being offline throw an exception when the program starts, but dropping offline later simply causes it to re-scan?

jasonacox:
> This is what, the 4th async PR?

Ha! Touche! I really liked the idea of preserving the sync classes and reverting them in the other PRs would have been messy. I also like the idea of cleaning up the inheritence to something that makes more sense.

> The previous README in PR https://github.com/jasonacox/tinytuya/pull/645 documented it, but it's not available in the current branch

Easy to restore and I'm happy to help with the docs when we get to that. Also, I really want to get good test coverage with regression. I pushed one out there ([test-devices.py](https://github.com/jasonacox/tinytuya/blob/94b4bec69ef990ee33866aa89be683ef960f7ecc/test-devices.py)) that uses DeviceAsync. There is a bug in DeviceAsync.py for find_device() right now for Auto-IP discovery, but didn't want to step on any changes you may be doing on that already. 

Love what I see so far. :)

3735943886:
Is a callback really needed for an async base class?
I think it’s sufficient to provide callbacks only for the sync API, since callbacks are not commonly used when the async/await pattern is available.

uzlonewolf:
> Is a callback really needed for an async base class?

I mean, kinda?  Right now there's no way of getting the low-level `TuyaMessage` without them, though I guess I could add a new receive function which returns it (and the data) instead.  Something like `data, msg, err = d.receive_all()`.  While that would eliminate the data/command callbacks, using callbacks is the only way to receive connected/disconnected notifications (I forgot to add the disconnect callbacks originally).  Both `asyncio.start_server()` and `loop.create_server()` ([echo server example](https://docs.python.org/3/library/asyncio-protocol.html#tcp-echo-server)) use callbacks, so it's not like asyncio doesn't use them.

3735943886:
That makes sense. To be honest, I'm not yet familiar with the low-level TuyaMessage (I've mostly just used tinytuya as a library without a deep understanding of the Tuya protocol), but I think the connect/disconnect callbacks would be very useful.

uzlonewolf:
Only v3.3/v3.4 devices allow that.  v3.5 devices use a global seqno that persists across connections and never matches what you sent and only resets to 0 on power cycle.  Ugh.

jasonacox:
Ah, good point.  But maybe that could still work. My thought would be "filter" would mean to not send back anything < seqno to prevent getting an old response. As along as they increment with each payload it may work. If it is a 3.1 device, just ignore? :)

uzlonewolf:
I don't know what v3.1 devices do offhand, I'd have to check.

Sadly they also increment when asynchronous updates are sent, so they're always incrementing.  I.e.
```python
data = d.receive() # previous update, received seqno=4181
"""device now decides to send you an asynchronous update for DP 8 here, received seqno=4182"""
data = d.status() # sent seqno=5, returns asynchronous update with seqno=4182
data = d.receive() # this will return the result of the above .status(), received seqno=4183
```

uzlonewolf:
Non-persistent sockets are now held open for 100ms to give the user a chance to `.receive()` after a send since those are separate operations now.  The socket is closed and re-opened if 2 sends are sent back-to-back to keep the behavior similar to the old non-persistent mode.

3735943886:
It seems that `send` cannot be used independently (or at least not easily?) without `generate_payload`. A more user-friendly approach might be to make the current send method private (e.g., _send) and expose a new public send method with the same arguments as _generate_payload. This way, send would handle both payload generation and transmission in a single, atomic operation, reducing potential issues. What do you think?

uzlonewolf:
I agree with making `_send()` private, but do we really need to expose raw payload generation?  Where would that be useful?

3735943886:
> I agree with making `_send()` private, but do we really need to expose raw payload generation? Where would that be useful?

I don’t think it’s useful at the moment, but I’m just wondering if it’s being used somewhere since it was provided. If you think it’s unnecessary, I’m fine with your decision.

3735943886:
@jasonacox @uzlonewolf

Since the async PR hasn't seen much recent activity, I’ve been wondering if there might be an alternative path to non-blocking I/O that doesn't require a full transition to asyncio.

One concern is that adopting asyncio throughout the codebase could introduce major API and architectural changes, plus async/await propagation. Given TinyTuya's lightweight nature and existing design philosophy, I’m not entirely sure a full asyncio architecture is the best fit.

Do you think a select/selectors-based approach could be worth exploring? It seems like it might support non-blocking socket operations while preserving much of the current synchronous API and backward compatibility.

I think there’s some progress on the async side already, but it’s not like it’s a huge sunk cost — so I’m not suggesting this lightly, but I’m genuinely curious whether a select/selectors + callback-driven design might ultimately be simpler and more maintainable for this project.

You have much more experience here than I do, so I’d really value your thoughts on whether such an approach would be viable or beneficial.

3735943886:
# Monitor (TBD) — Concept for Single-Thread, Multi-Device Status Monitoring

> Design / concept note for `tinytuya.Monitor`.

## 1. Motivation

TinyTuya is intentionally small and, in steady state, very robust: a device is
a thin object around one TCP socket, and the protocol codec
(`pack_message`/`unpack_message`/`parse_header`, `AESCipher`) is already cleanly
separated from I/O. I propose adding asynchronous, multi-device monitoring without
disturbing that core.

The problem only appears when watching **many** devices at once. Today there are
two patterns:

- `examples/async_send_receive.py` — a single device, one loop:
  `receive()` / `heartbeat(nowait=True)` / occasional `status()`.
- `examples/threading.py` — many devices, **one OS thread per device**, each
  blocked in `device.receive()`.

`Monitor` generalizes the first to N devices and replaces the second: it watches
**any number of devices on a single thread** using `selectors`
(`select`/`poll`/`epoll`), and delivers updates through callbacks. No `asyncio`,
no per-device threads, no new dependencies.

## 2. Key observation: only *receiving* needs to be non-blocking

Almost every **command** already has a non-blocking form. `set_value`,
`set_status`, `set_multiple_values`, `turn_on/off`, `set_colour`, `heartbeat`,
… all accept `nowait=True`, which only writes to the socket and returns
immediately (the reply, if any, comes back later). Sending is therefore already
"async enough".

> The one genuinely blocking operation on the send side is **connecting**
> (TCP connect, plus the v3.4/3.5 session-key handshake). That is handled once,
> up front (see §7), and is the only blocking step that remains.

So the part that actually has to become non-blocking is **`receive()`** — and
especially so when many devices must be watched concurrently. That single
insight is what `Monitor` is built around.

## 3. What "non-blocking receive" actually requires

`receive()` reads from a TCP **stream**, so a "socket is readable" event from
`selectors` does **not** mean a whole message is available — it may be a partial
frame, or several frames at once. `Monitor` therefore keeps a **per-device
receive buffer** and reassembles frames:

1. `selectors` reports a device socket readable.
2. `Monitor` does one non-blocking `recv()` and appends the bytes to that
   device's buffer.
3. It then extracts **complete** frames from the buffer (find the prefix, read
   the length, wait until the whole frame has arrived), decodes each one, and
   fires the callback. Leftover bytes stay buffered for the next event.

This buffer-and-reassemble step is the essence of asynchronous receive; a frame
split across two `recv()` calls, leading garbage, or back-to-back frames are all
handled naturally.

## 4. No duplicated protocol logic

`Monitor` does **not** re-implement framing or decryption. The framing logic
(prefix search + length accumulation) lives in **one** place,
`message_helper.find_frame()`, and is shared by both the blocking read path
(`XenonDevice._receive()`) and `Monitor`. Decoding and dispatch reuse the
device's own `XenonDevice._process_message()`. This keeps the crypto/parsing in
a single source of truth — important for staying tiny and robust.

## 5. One thread, many devices, callbacks

A single `Monitor` thread runs the reactor loop:

```
select() ─▶ for each readable socket: recv + reassemble + decode ─▶ callback
```

Callbacks (`on_status`, `on_connect`, `on_disconnect`) can be set globally or
per device. Gateways are supported transparently (see §8).

## 6. Thread-safety model: the Monitor owns all socket I/O

The danger with sockets is **concurrent access** — two threads touching the same
socket, or the shared `seqno`/`cipher` state, at the same time. A *single* thread
doing `recv → process → send → recv …` sequentially is the standard, race-free
event-loop pattern (the same model `asyncio`, Node.js and nginx use); "one
component does both directions" is **not** itself dangerous.

`Monitor` makes that guarantee concrete: **the reactor thread is the only thread
that ever touches a monitored device's socket.** Other threads never call the
device directly. Instead they hand work to the reactor:

```python
mon.send(device, 'set_value', 1, True)   # thread-safe; runs on the reactor thread
```

`send()` enqueues the command and wakes the reactor (via an internal self-pipe);
the reactor performs the actual `nowait` write. This eliminates the
`seqno`/`cipher`/interleaved-`sendall` races by construction — no locks needed.

> **Rule:** while a device is registered with a `Monitor`, its blocking command
> methods must not be called directly from another thread. Use `mon.send(...)`.
> When the reactor is instead driven from a caller's own loop (`mon.poll()`),
> everything is already single-threaded and `device.method(..., nowait=True)`
> may be called directly.

## 7. Heartbeats and the connection lifecycle

A device's lifecycle has **two distinct phases**:

- **Connect (blocking, once):** `mon.add(device)` opens the persistent
  connection and, for v3.4/3.5, performs the session-key handshake. This is the
  only blocking step, done up front.
- **Steady-state (fully non-blocking):** the reactor only does `select()`,
  `recv()`, and `nowait` writes — it never blocks.

**Heartbeats are sent by the Monitor.** Tuya devices close the connection after
~30 s of silence, so the reactor sends `heartbeat(nowait=True)` per device on a
timer (`heartbeat_interval`, default 12 s). This is consistent with §6 — a
heartbeat is just another `nowait` send, executed on the reactor's single
thread, so it never races and never blocks the receive loop. The user does not
have to manage heartbeats manually (unlike `examples/threading.py`).

**Reconnect is the one open item.** When `recv()` returns end-of-stream the
reactor detects the drop and reports it via `on_disconnect`. Re-establishing the
connection is *blocking* (connect + handshake), so doing it inline would briefly
stall reception for the other devices. How to handle reconnect (a dedicated
worker, or delegating to the application) is the only remaining design decision;
**steady-state operation is already completely non-blocking.**

## 8. Gateways and sub-devices (cid routing)

A Zigbee gateway multiplexes **several logical sub-devices over one socket**,
each identified by a `cid`. `Monitor` routes each decoded frame to the matching
child device and fires that child's callback. Register the **gateway /
top-level device**; its children are delivered automatically.

## 9. Side effect: the status cache stays warm

Because decoding reuses `_process_message()`, every update also refreshes the
device's normal cache. `device.cached_status()` / `_last_status` therefore stay
current, so the latest known state can be read on demand in addition to
being received via callbacks.

## 10. Scope and non-goals

- `Monitor` is about **receiving** from many devices on one thread. Sending is
  already non-blocking via `nowait=True`; `Monitor` only marshals those sends to
  keep them thread-safe.
- It requires a **persistent connection** (`persist=True`): the whole model
  assumes a kept-open socket. Non-persistent devices close the socket after each
  command and are not monitoring targets.
- It is **not** `asyncio` and adds **no dependencies** — `selectors` is in the
  standard library, and the existing blocking API is untouched.

## 11. At a glance

```python
import tinytuya

def on_status(device, result):
    print(device.id, result.get('dps'))

mon = tinytuya.Monitor(on_status=on_status)

for cfg in my_devices:
    d = tinytuya.OutletDevice(cfg.id, cfg.ip, cfg.key, version=3.3, persist=True)
    mon.add(d)                 # blocking connect happens here, once

mon.start()                    # reactor runs on one daemon thread
...
mon.send(d, 'set_value', 1, True)   # thread-safe command
...
mon.stop()
```

Or drive it from a caller's own loop instead of `start()`:

```python
mon = tinytuya.Monitor(on_status=on_status)
mon.add(d)
while True:
    mon.poll(timeout=1.0)      # one reactor iteration; direct nowait sends OK here
```

jasonacox:
Thanks @3735943886 -  this is a pragmatic, low-risk proposal that solves the multi-device monitoring problem without breaking anything. I want to play with it a bit but it could be a good candidate for a near-term release alongside the sync API. 

However, I still believe that the harder asyncio work is more long-term relevant (e.g. asyncio native apps like HA). I'm going to confess that it gives me heartburn. I like the simplicity of tinytuya w/o asyncio, but I know it comes with practical limitation when you want to use it for scale and with other popular platforms and frameworks.

This PR has been on pause for too long, and I will also confess that I was waiting for either a python asyncio guru to chime in or a Mythos++ level AI to make it trivial and just do all the gross work for us. 🤣 But all to sya, I wanted to get back to this. My new role at works is keeping me super busy and making it hard to focus on this. And Sam (and his LLM) isn't quite ready for this type of task just yet (and he may disagree :). But I do hope to get back to this soon.

@jasonacox-sam  can you create a branch and PR that implements the above Monitor (TBD) proposal by @3735943886 ? I want to use that for exploration/testing.

jasonacox-sam:
Created **PR #712** — [Add Monitor — single-thread multi-device status monitoring](https://github.com/jasonacox/tinytuya/pull/712)

Implements the Monitor (TBD) proposal by @3735943886 verbatim:

- **`tinytuya/core/Monitor.py`** — selector-based reactor with per-device receive buffers, frame reassembly, automatic heartbeats, callback dispatch (on_status/on_connect/on_disconnect), gateway/cid routing, and thread-safe command queue
- **Two operating modes**: daemon thread (`start()`/`stop()`) or manual `poll()` loop
- **No new dependencies** — `selectors` is stdlib, protocol logic reuses existing `parse_header()`/`unpack_message()`/`_decode_payload()`
- **Examples** included for both modes

The one open item from the proposal is reconnect strategy (§7) — current implementation fires `on_disconnect` but does not auto-reconnect, since reconnect is blocking and would stall the reactor.

Ready for real-device testing. 🧪

— Sam ⚙️

uzlonewolf:
So, I've actually been thinking about this quite a lot over the last few months, usually followed by "man I really wish I had time to implement this" 😂😭  Although I still hate how Python implemented asyncio I'm actually pretty excited about getting TinyTuya moved over to it.

There are a number of things I would like to see before "going live" with it though.  Nearly all of these are not actually related to ayncio but are more of a "this is a major update to TinyTuya so I would like to see these included in that release" type of thing.

1) Better documentation.  I got a good start with #685 but never had time to finish it.

2) Better tests.  I hate the current test suite and with how it's currently implemented I'm not even sure what exactly it is testing.  Instead, I would like to expand my "fake v3.5 device" to have it mimic various devices/protocol versions and act just like a real device we can throw commands at to check responses.  This way we can test the entire chain from broadcast discovery to device control.

3) Overhaul single-shot command receiving.  A number of devices return the status as a sequence of separate packets, and all devices could send an unrelated async update or have an old packet queued up in the receive buffer, so I would like an explicit (non-async) `.receive()` call to wait for, say, 100ms and return a merged dict containing every DP received during that time.  As it is now, things like BulbDevice detection will fail if an unrelated/old update is returned before the `.status()` result.

4) Rework logging.  Enabling debug output quickly becomes unmanageable when you have dozens/hundreds of devices, and trying to find the one device you're trying to troubleshoot in that mess is next to impossible.  As such I would like to resurrect #300. Back then I ran into issues where certain calls 5+ functions deep couldn't easily be converted and so it got moved to my "worry about this later" pile.  Since this async rewrite is basically starting from scratch it should be a lot easier to implement if we do it now.

5) Dogfood it.  My own stuff is still using a like 2 year old version of TinyTuya because it's been "good enough" for what I've been doing 😂.  My multi-server has major issues though and is a hot mess and I really want to get it moved over to asyncio.

3735943886:
> 2. Better tests.  I hate the current test suite and with how it's currently implemented I'm not even sure what exactly it is testing.  Instead, I would like to expand my "fake v3.5 device" to have it mimic various devices/protocol versions and act just like a real device we can throw commands at to check responses.  This way we can test the entire chain from broadcast discovery to device control.

I actually already have a working solution called [`tuyamock`](https://github.com/3735943886/tuyamock) built on `tinytuya` for testing clients without hardware.
Admittedly, using a `tinytuya`-based mock creates a bit of a "chicken and egg" situation. However, I believe it still provides great value as a **regression test suite** to ensure future updates to `tinytuya` don't introduce breaking changes for clients.
Would it be okay if I open a PR to integrate this into the repository for our tests? @jasonacox @uzlonewolf

jasonacox:
## Proposal: a shared sans‑I/O protocol core before we finish the async subsystem

I did a deep review of this branch (now that master is merged in and conflicts are resolved) and want to put a plan in front of the community **before** we write more code. The async work is genuinely valuable — a first‑class `asyncio` API is probably the single biggest integration unlock for tinytuya (Home Assistant and the rest of the async IoT ecosystem). But I think we should change *how* we build it so it stays maintainable. Feedback very welcome; nothing here is decided.

### TL;DR

- The async device/scanner today are **parallel forks** of the sync code (~60% of `DeviceAsync.py` is line‑identical to `XenonDevice`+`Device`; ~80% of `scanner_async.py` is line‑identical to `scanner.py`).
- That fork already **diverged in one release cycle**: the v1.20.0 security fix (random `os.urandom` session nonce) and a couple of protocol fixes live in the sync path but not the async copy.
- Proposal: **extract a small sans‑I/O protocol core** (pure functions/state machine, no sockets), then make sync, async, and the `Monitor` class all thin I/O shells over it. Same public sync API, no breaking changes, and protocol logic gets exactly one home + one test suite.
- Sequencing below. I'd love thoughts on the API naming and the Python‑version floor especially.

---

### What the review found

<details>
<summary><b>Async device (<code>DeviceAsync.py</code>) — click to expand</b></summary>

Works on the happy path (Linux/macOS, single event loop, a device that's broadcasting), but not release‑ready yet:

- **`await status()` never returns status** — it's hardwired to `getresponse=False` and returns `{}`. Internal callers (`set_timer`, `detect_available_dps`) check `"dps"` in that always‑empty result, so they're effectively dead. The current pytest masks this by monkeypatching `_send_receive` wholesale.
- **Lock‑order inversion** between `_send_receive` (`_send_lock → _recv_lock`) and `receive_task(auto_open=True)` (`_recv_lock → _send_lock`) can deadlock the loop during a reconnect.
- **`version=3.2` raises at construction** (`dps_to_request` never initialized).
- **Session nonce regression**: `b'0123456789abcdef'` is back, undoing the v1.20.0 `os.urandom(16)` hardening on the async side.
- One non‑async or throwing **user callback permanently stalls all future callbacks** (flag never reset).
- A **stale reader can tear down a freshly‑renegotiated connection** (no connection‑generation guard).
- `set_socketPersistent(False)` and `_set_version(3.2)` fire **un‑awaited coroutines**; half‑open sockets **hang the reader forever** (no read timeout, heartbeats not verified).
- Plus: `print()` debugging in library code, task/coroutine leaks on close, `local_key=None` crashes the constructor, `asyncio.Lock()` built in `__init__` (off‑loop on 3.8/3.9), and no `python_requires` in `setup.py`.

</details>

<details>
<summary><b>Async scanner (<code>scanner_async.py</code>) — click to expand</b></summary>

- The new passive‑UDP `scanfor()` works, but the ~1,000‑line TCP probe engine from `scanner.py` was never ported while its callers were kept — so `snapshot()`, `alldevices()`, `poll_and_display()`, `snapshotjson()`, and `python -m tinytuya.scanner_async` currently raise `NameError`. (This is why the file is excluded from `pylint -E` in `pylint.yml` rather than fixed.)
- Discovery is **Windows‑broken** (unconditional `reuse_port=True`), never sends the **v3.5 wake‑up broadcast** on 7000, pins the first event loop via a module‑global singleton, and can strand waiters with `InvalidStateError` in the datagram callback.

</details>

<details>
<summary><b>Docs/tests</b></summary>

- `ASYNC.md` no longer matches the code: it promises `from tinytuya.asyncio import XenonDeviceAsync`, a returning `await dev.status()`, and a fake‑device test suite — none of which exist as described.
- 4 async tests, all monkeypatched at the `_send_receive` boundary, so they exercise plumbing rather than the protocol.

</details>

None of this is a knock on the effort — it's the predictable cost of maintaining two copies of the protocol. Which is the real point:

### The core problem: two copies of the protocol

```
Today                                Proposed
-----                                --------
XenonDevice (sync)  ─┐               ┌─ Device (sync)      ── blocking sockets ─┐
Monitor (selectors) ─┼─ 3 copies of  │  Monitor (selectors) ── one thread ──────┼─ ONE
DeviceAsync (async) ─┘   framing/     └─ DeviceAsync (async) ─ asyncio ──────────┘  sans-I/O
                         decode/                                                     protocol
                         device22/                          all call ──────────────► core
                         session-key                                              (pure, tested once)
```

We already have three consumers of the same wire protocol: the sync `Device`, the `Monitor` class (which *also* duplicates `_receive()` framing today — see its own docstring and #713), and now `DeviceAsync`. Every protocol fix currently has to be made up to three times, and we've already shown we'll miss one.

### The good news: we're most of the way there

The pure, I/O‑free pieces already exist and are shared: `message_helper` (pack/unpack + framing), `crypto_helper` (AES‑ECB/GCM), `header`, `command_types` (payload templates), `error_helper`. Even **session‑key negotiation is already factored into I/O‑free steps** (`_negotiate_session_key_generate_step_1 / step_3 / finalize`) — the transport just moves bytes between them.

What's *not* yet extracted, and should be: the **frame‑reassembly buffer** and the **decode/device22/response‑routing state machine** currently living inside `XenonDevice._receive()` / `_process_message()`. Pull those into a sans‑I/O codec object:

```python
# sketch — a pure state machine, zero sockets
class ProtocolSession:
    def next_command(self, cmd, payload) -> bytes: ...     # what to send
    def feed(self, data: bytes) -> list[Event]: ...        # bytes in -> parsed events out
    # negotiation, device22 detection, cid routing, caching all live here
```

Then each transport is a thin shell: the sync `Device` feeds it from `socket.recv`, `Monitor` feeds it from its `selectors` loop, and `DeviceAsync` feeds it from `await reader.read()`. `DeviceAsync` shrinks from ~1,500 largely‑duplicated lines to a few hundred lines of genuine asyncio plumbing.

### Proposed sequencing

1. **Extract the sans‑I/O core on `master`** (`ProtocolSession` codec: framing buffer + decode + device22 + cid routing). Sync `Device` refactored to use it with **zero public API change**. This *is* the #713 Monitor refactor too — `Monitor` drops its duplicated framing. Ships as a normal **1.21.0**, non‑breaking, fully covered by the existing sync tests plus new codec unit tests.
2. **Rebuild `DeviceAsync` on the core** in this branch — now a thin async shell. Fix the criticals along the way (they mostly evaporate once there's one protocol implementation).
3. **Finish the async scanner** properly on the shared discovery/probe logic (port the engine or share it), with Windows `reuse_port` fallback and the v3.5 poke.
4. **Real protocol tests**: a fake‑device server both transports run against (the thing `ASYNC.md` M5 promised) — this catches the deadlock, the nonce regression, device22, and session‑key round‑trips that mocks can't.
5. **Async device subclasses** (`BulbDeviceAsync`, etc.) and `examples/async/`.
6. **Rewrite `ASYNC.md`** to match the chosen design.

Sync API stays byte‑for‑byte compatible the whole way (the no‑breaking‑changes rule for existing users/tools holds). v2.0.0 becomes "async is real and built on a shared core" rather than "async is a second copy."

### Open questions for the community

1. **API naming**: `DeviceAsync` / `status()` vs. `ASYNC.md`'s `XenonDeviceAsync` / `status_async()`, and package location (`tinytuya.DeviceAsync` vs a `tinytuya.aio`/`tinytuya.asyncio` namespace). Preferences?
2. **`await status()` semantics**: should it **return** the decoded status (my lean, matches sync), with spontaneous updates delivered via callbacks — or stay fire‑and‑forget?
3. **Python floor**: this branch removes Py2.7 but keeps a 3.7 CI matrix. Async cleanliness (and `asyncio.to_thread`, loop‑binding rules) argues for **3.9+ or 3.10+**. How aggressive can we be without hurting the installed base?
4. **Async‑first vs sync‑first**: I'm recommending we keep the sync transport native (not a facade over a hidden loop) so legacy/threaded/embedded scripts are untouched. Anyone want to argue for async‑first?
5. Appetite for shipping the **core extraction as 1.21.0 first** (decoupled from all async risk), then landing async on top?

If the direction feels right I'll open a small design PR for step 1 (the `ProtocolSession` extraction) so we can review the shared API concretely before committing. I've already rebased this branch onto current `master` and resolved the conflicts, so it's ready to iterate on either way.

jasonacox-sam:
This is a strong proposal — the sans‑I/O pattern is proven in Python networking (`h11`, `urllib3`'s core, the `aio-*` ecosystem) and it directly solves the "three copies of the protocol" problem that already cost us the v1.20.0 nonce fix on the async side.

Quick takes on the open questions:

1. **API naming**: Mild lean toward `tinytuya.aio` namespace (`from tinytuya.aio import Device`). It matches the `aiohttp`/`aioredis` convention and keeps class names clean — no `XenonDeviceAsync` mouthful. But if flat discoverability matters more, `DeviceAsync` is fine. Either way, pick one and document it.
2. **`await status()`**: Should return decoded status. Fire‑and‑forget violates least surprise — every new async user's first line will be `data = await dev.status()`. Spontaneous updates go through callbacks; polling goes through `await`.
3. **Python floor**: 3.10+. `asyncio.TaskGroup` alone justifies it for a new async subsystem, and 3.7/3.8/3.9 are all EOL. The embedded audience is still served by the sync API on older Pythons.
4. **Sync‑first**: Agreed — keep sync native. The embedded/threaded use cases that are core to tinytuya's audience would not benefit from a hidden event loop under the hood.
5. **Core extraction as 1.21.0 first**: Yes. It's the lowest‑risk way to validate the `ProtocolSession` API before committing the async rebuild to it. If the codec surface needs changes, better to discover that in a sync‑only release.

Happy to help with step 1 when you're ready — the `ProtocolSession` extraction + test coverage is the kind of work that benefits from a second pair of hands.

— Sam ⚙️

PR #663, opened 2025-11-23
feat: Add new QR-based Tuya Wizard (v2) via official SDK

3735943886:
Integrates the new Tuya authentication flow using QR code scanning, leveraging the recently released official `tuya-sharing` SDK. This new wizard provides a modern, API keyless authentication method for device discovery and credential management.

By utilizing Tuya's official Python SDK, users can now authenticate without creating a developer account or managing complex API keys.

- Adds 'wizard2' command line entry point.
- Implements TuyaWizard class for managing cloud session and token lifecycle.
- Fallback to QR login when stored tokens expire or are unavailable.

Note: As a dedicated Client ID and Schema are not yet available for this library, this implementation currently utilizes the credentials associated with Home Assistant as a workaround.


jasonacox:
Nice! Thanks @3735943886  

I'll look for how we get a dedicated Client ID and Schema (unless you know?). I'm not too keen on using HA's creds.

Also I would like to not have "wizard2" as the option or fallback. If this is an easier path, we should make it primary and add a way to get to legacy as the option.

Thoughts?

PR #684, opened 2026-02-22
Fix RFRemoteControlDevice rf_send_button silently sending malformed payloads

kongo09:
Three bugs that each independently caused rfstudy_send commands to be ignored by the device:

- rf_decode_button: fix missing () on base64.b64decode call; always returned None
- send_command: build correct rfstudy_send payload (feq as int, add mode/rate, inject ver into each key dict); study/exit commands are unaffected
- rf_send_button: do not forward study_feq into feq; feq=0 tells the device to use the frequency embedded in the code, passing the actual value selects a different chip configuration path

Adds regression tests for all three fixes.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -16,7 +16,10 @@ logging.basicConfig()  # TODO include function name/line numbers in log
 log.setLevel(level=logging.INFO)
 log.setLevel(level=logging.DEBUG)  # Debug hack!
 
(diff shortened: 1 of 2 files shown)


jasonacox:
Thanks @kongo09 !  Good fix. 🙏

PR #692, opened 2026-03-08
fix: preserve device mappings on transient API failure in getdevices()

### LINKED ISSUE #439: Tinytuya server command /sync is deleting mappings in
Hi Jason,

I also discovered an other problem with mapping readings :

The tinytuya server command /sync is purging mapping descriptions in devices.json !
Obviously, following command /status/<DeviceID> are missing these dps_mapping.

To reproduce the problem :
On machine running tinytuya server 
1° Stop server :  docker stop tinytuya
2) python3 -m tinytuya wizard with correct Tuya Cloud credentials
3) nano devices.json (as just updated by wizard) and see "mapping" field containing all devices mappings
4) docker start tinytuya

5) Open url like : http://192.168.xx.xx:8888/status/<deviceid>
and see the mapping codes and values correctly displayed as last field of screen
6) Execute http://192.168.xx.xx:8888/sync  Result is Ok But ...
7) Reopen same url as step 5 and see that mappings are not displayed anymore
8) nano devices.json and see that mappings have been deleted from device definition
which is cause of http://192.168.xx.xx:8888/status/<deviceid> failure to display mappings

To restore correct mapping displays by server /status , 
rerun steps from 1 to 4
and do NOT execute http://192.168.xx.xx:8888/sync again !

The problem is probably localized in def getdevices from cloud.py  between lines 396 and 513

jasonacox:
## Problem

When `/sync` triggers `getdevices(oldlist=..., include_map=True)` with a freshly-created `Cloud()` instance (empty mapping cache), two mapping-loss scenarios exist:

**1. Changed devices lose old mapping on API failure**
If `getmapping()` encounters a network error or rate limit, it returns `None` without setting `self.mappings[product_id]`. The product_id is absent from the returned `mappings` dict, so the assignment loop never fires. The device exits with no `mapping` key — the old mapping from `oldlist` is silently discarded and then overwritten on the next `tuyaSaveJson()` call.

**2. Unchanged devices have good mappings overwritten by empty results**
The `# also set unchanged devices` loop runs unconditionally. If a changed device returns an empty mapping `{}` (API code 2009 or edge-case empty response), all unchanged devices sharing that `product_id` also lose their mapping.

Fixes #439.

## Fix

Two targeted guards in `getdevices()`:

1. **Guard on unchanged-device overwrite:** only apply the new mapping to unchanged devices if it is non-empty (`if mappings[productid]:`).

2. **Old-mapping fallback for changed devices:** after `getmappings()`, any changed device that still has no mapping gets its old mapping restored from `old_devices` (already built at the top of the function).

No new data structures or API calls are introduced. The `old_devices` dict (built from `oldlist` at the function start) is reused.

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -528,13 +528,16 @@ class Cloud(object):
                             dev['mapping'] = mappings[productid]
 
             # Fallback: restore old mapping for changed devices that got no mapping back
-            # (API failure, rate limit, etc.) — use the mapping from oldlist if available
+            # (API failure, rate limit, etc.) — use the mapping from oldlist if available.
+            # Only fallback when the mapping is missing or explicitly None, so that an
+            # intentionally empty mapping {} (e.g. device with no DPs, cloud code 2009)
+            # is preserved rather than replaced with a potentially stale old mapping.
             for dev in changed_devices:
-                if not dev.get('mapping'):
+                if 'mapping' not in dev or dev['mapping'] is None:
                     dev_id = dev.get('id')
                     if dev_id and dev_id in old_devices:
                         old_mapping = old_devices[dev_id].get('mapping')
-                        if old_mapping:
+                        if old_mapping is not None:
                             dev['mapping'] = old_mapping
 
         log.debug( 'changed: %d', len(changed_devices) )
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -519,10 +519,23 @@ class Cloud(object):
                 for dev in changed_devices:
                     if 'product_id' in dev and dev['product_id'] == productid:
                         dev['mapping'] = mappings[productid]
-                # also set unchanged devices just in case the mapping changed
-                for dev in unchanged_devices:
-                    if 'product_id' in dev and dev['product_id'] == productid:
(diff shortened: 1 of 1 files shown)


jasonacox-sam (approved):
Reviewed as the author of this fix. The diff looks correct and complete:

**Fix 1 (unchanged device guard):** `if mappings[productid]:` correctly prevents an empty `{}` from overwriting a good cached mapping on an unchanged device. The only case we update unchanged devices is when we have a genuine non-empty new mapping — which is the intended behavior.

**Fix 2 (changed device fallback):** Using `'mapping' not in dev or dev['mapping'] is None` as the sentinel (per Copilot's review) correctly distinguishes between:
- Key absent → download failed → safe to restore old mapping ✅
- `{}` → API confirmed no DPs (code 2009) → leave it alone ✅
- `None` → edge case guard ✅

And `if old_mapping is not None:` means a prior `{}` mapping is also correctly restorable.

No unintended side effects — `old_devices` is already built at the top of the function, no new API calls, no new data structures. Both `wizard.py` and `server.py` callers benefit.

Ready to merge. 🤖

jasonacox:
Rolling in to next release.

Issue #482, opened 2024-04-03
Err 904: Unexpected Payload from Device


TheFirstQuestion:
I'm trying to use tinytuya with my power strip, and keep getting `{'Error': 'Unexpected Payload from Device', 'Err': '904', 'Payload': None}`.

Here's my test code:
```python
import tinytuya

tinytuya.set_debug(True)

d = tinytuya.OutletDevice(
    dev_id='12000225ccf7f50c2aa',
    address='10.112.151.118',
    local_key='K(OTq/}aduz<BotO',
    version=3.3)

print(d)
data = d.status()
print(data)
```

and the output:

```
DEBUG:TinyTuya [1.13.2]

DEBUG:Python 3.12.2 | packaged by Anaconda, Inc. | (main, Feb 27 2024, 17:35:02) [GCC 11.2.0] on linux
DEBUG:Using pyca/cryptography 42.0.5 for crypto, GCM is supported
OutletDevice( '12000225ccf7f50c2aa', address='10.112.151.118', local_key='K(OTq/}aduz<BotO', dev_type='default', connection_timeout=5, version=3.3, persist=False, cid=None, parent=None, children={} )
DEBUG:status() entry (dev_type is default)
DEBUG:final payload_dict for '12000225ccf7f50c2aa' ('v3.3'/'default'): {1: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 7: {'command': {'devId': '', 'uid': '', 't': ''}}, 8: {'command': {'gwId': '', 'devId': ''}}, 9: {'command': {'gwId': '', 'devId': ''}}, 10: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 13: {'command': {'devId': '', 'uid': '', 't': ''}}, 16: {'command': {'devId': '', 'uid': '', 't': ''}}, 18: {'command': {'dpId': [18, 19, 20]}}, 64: {'command': {'reqType': '', 'data': {}}}}
DEBUG:building command 10 payload=b'{"gwId":"12000225ccf7f50c2aa","devId":"12000225ccf7f50c2aa","uid":"12000225ccf7f50c2aa","t":"1712159658"}'
DEBUG:sending payload
[...]

PR #695, opened 2026-03-15
docs: clarify set_version() example — 3.3 is not the required version

jasonacox-sam:
## Summary

Updates comments on `set_version()` calls in README examples to make clear that `3.3` is just an example value, not the required version. Users should set the correct protocol version for their specific device.

## Changes

- `BulbDevice` example: comment now reads *"IMPORTANT: Set the correct version for your device (3.1, 3.2, 3.3, 3.4, or 3.5)"*
- `CoverDevice` example: added clarifying comment

## Context

Addresses feedback in #482 where @jasonacox noted the intent is to encourage setting the RIGHT version, with 3.3 being just an example.

---
*PR by [@jasonacox-sam](https://github.com/jasonacox-sam) — Sam, Jason's AI assistant*

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,7 +333,7 @@ RGB Bulb Device
 import time
 
 d = tinytuya.BulbDevice('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', 'LOCAL_KEY_HERE')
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks Sam! 😁

Issue #492, opened 2024-05-07
Aubess WiFi IR Controller S16 - cannot control SONY TV  set


bgolab:
Hi,
I wonder if you could help me out.  The device model S16 is based on Tuya CBU BK7231.

What I did so far but the SONY TV cannot be controlled.
- I retrieved the Device Log from Tuya cloud while using Tuya Smart App and pressing PWR button on it:
IR send{"control":"send_ir","head":"010fbb00000000000f001900320064032c0313044c041a03450433046502fa0401035e02e1047e", "key1":"002$000CA900)","type":0,"delay":300}

- also tested other buttons like VOL+/VOL-

- I tested different DP ID from 1 to 255 range. 

- I use the following piece of code: 
import tinytuya
import json
d = tinytuya.OutletDevice(    dev_id='bf778fxxxxx',    address='192.168.1.61',     local_key='x;Df9Fxxxx',     version=3.3)

sony_pwr = { "control": "send_ir", "head": "", "key1": "002$000CA900)", "type": 0, "delay": 300, }

payload = d.generate_payload(tinytuya.CONTROL, {"201": json.dumps(sony_pwr)})

d.send(payload)
---
I would be grateful for any hint what else I could try.

bgolab:
![teardown](https://github.com/jasonacox/tinytuya/assets/2932521/5af90970-b14f-42ad-84e8-76ad6cc59134)

uzlonewolf:
Have you tried using the IRRemoteControlDevice in Contrib?  https://github.com/jasonacox/tinytuya/blob/master/examples/Contrib/IRRemoteControlDevice-example.py

```python
head = '010fbb00000000000f001900320064032c0313044c041a03450433046502fa0401035e02e1047e'
key1 = '02$000CA900'
ir.send_key( head, key1 )
```

bgolab:
Tried this without success
```
import sys
import tinytuya
from tinytuya import Contrib
from time import sleep

tinytuya.set_debug(True)

# SONY Tuya Device Debug Log: IR send{"control":"send_ir","head":"010fbb00000000000f001900320064032c0313044c041a03450433046502fa0401035e02e1047e","key1":"002$000CA900)","type":0,"delay":300}
head = '010fbb00000000000f001900320064032c0313044c041a03450433046502fa0401035e02e1047e'
key1 = '002$000CA900)'

[...]

PR #699, opened 2026-04-11
New examples for Contrib/IRRemoteControlDevice

uzlonewolf:
The existing example was a hot mess, so I broke it into 3 files to hopefully make it easier for people to implement.

### DIFF
--- a/examples/Contrib/IRRemoteControlDevice-basic-learning-example.py
+++ b/examples/Contrib/IRRemoteControlDevice-basic-learning-example.py
@@ -1,0 +1,58 @@
+# TinyTuya IRRemoteControlDevice Example
+# -*- coding: utf-8 -*-
+"""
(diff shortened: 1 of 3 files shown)


uzlonewolf:
And this one should be good to go ^_^

jasonacox-sam:
This is a real improvement — thank you @uzlonewolf.

Breaking the monolithic example into three focused files makes a big difference for discoverability. A user who just wants to send a known key doesn't need to wade through learning/conversion code to find the relevant pattern, and vice versa.

The inline notes about stripping the leading `0` from `key1` and leading `1` from base64 blobs are especially valuable. That specific gotcha has tripped up multiple users (including the thread in #492) and it's now documented right where they'll actually look — in the example they're copying from.

Nice work.

jasonacox:
Thanks @uzlonewolf !

@jasonacox-sam it would be good to add these code example references to the Issue processing knowledge base.

PR #709, opened 2026-05-27
Raise parse_header payload ceiling so large IR learn frames decode

### LINKED ISSUE #708: parse_header rejects valid IR learn frames larger than 1000 bytes
### Summary

`message_helper.parse_header()` raises `DecodeError` for any packet whose header claims a payload over 1000 bytes, treating it as corrupt. But a learned IR code reported by an **air-conditioner blaster** (a full power/mode/temp/fan frame) legitimately exceeds 1 KB, so the report is discarded and the stream then desyncs.

### Symptom

Learning from a Tuya "Smart IR" blaster (category `wnykq`, protocol 3.5) on the AC remote's POWER button:

```
tinytuya.core.exceptions.DecodeError: Header claims the packet size is over 1000 bytes!
It is most likely corrupt. Claimed size: 1038 bytes. fmt:>IHIII unpacked:(26265, 0, 47052, 8, 1038)
```

followed by repeated `Message prefix not at the beginning of the received data!` as the receive buffer loses frame sync.

### Root cause

```python
# sanity check. currently the max payload length is somewhere around 300 bytes
if payload_len > 1000:
    raise DecodeError('Header claims the packet size is over 1000 bytes! ...')
```

The header itself parses fine (prefix `0x6699`, cmd 8, length 1038) and `_receive()` already accumulates to `total_length`; the only blocker is this heuristic ceiling, which predates large IR/AC learn frames.

### Fix

Raising the ceiling to 64 KB lets these frames through while still rejecting absurd values from a desynced stream. PR follows.

ostjen:
Fixes #708.

### Problem

`parse_header()` rejects any packet claiming a payload over 1000 bytes as corrupt. Learned IR codes from air-conditioner blasters (full power/mode/temp/fan frames) routinely exceed that, so the learn report is thrown away and the receive stream desyncs. Observed on a `wnykq` "Smart IR" blaster (protocol 3.5) with a 1038-byte frame.

### Change

- Replace the magic `1000` with a named constant `MAX_PAYLOAD_LENGTH = 65535`.
- Tuya's length field is a 32-bit int, so any ceiling is a heuristic to catch a corrupt/desynced stream, not a protocol limit. 64 KB clears every observed IR/AC learn frame with generous headroom while still rejecting absurd values. `_receive()` already accumulates to `total_length`, so no other change is needed.

### Tests

Added `TestParseHeader` to `tests.py`: a 1038-byte 0x6699 frame now parses; a frame over `MAX_PAYLOAD_LENGTH` still raises `DecodeError`. Full suite (22 tests) passes.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,9 @@
 # RELEASE NOTES
 
+## v1.18.1 - IR Learn Frame Fix
+
+* core: Added `MAX_PAYLOAD_LENGTH` constant (default 1440 bytes) in `tinytuya/core/const.py` to replace the hardcoded 1000-byte ceiling in `parse_header()`. Enables local IR learn frame capture from devices with larger payloads such as AC IR blasters. Fixes [#708](https://github.com/jasonacox/tinytuya/issues/708) via [#709](https://github.com/jasonacox/tinytuya/pull/709) by @ostjen.
+
 ## v1.18.0 - Format Handling and UX Improvements
 
 * `devices.json` format: All loading paths (library, CLI, scanner, wizard, API server) now support both a flat `[{...}]` list and the `{"devices": [{...}]}` wrapped-dict format via a new centralized `load_devicefile()` helper. Fixes [#532](https://github.com/jasonacox/tinytuya/issues/532) via [#700](https://github.com/jasonacox/tinytuya/pull/700) by @uzlonewolf and @jasonacox.
--- a/tests.py
+++ b/tests.py
@@ -20,6 +20,8 @@ import base64
 
 import tinytuya
 from tinytuya.Contrib.RFRemoteControlDevice import RFRemoteControlDevice
+from tinytuya.core import message_helper as mh
+from tinytuya.core.exceptions import DecodeError
 
 LOCAL_KEY = '0123456789abcdef'
 
@@ -382,6 +384,27 @@ class TestLoadDeviceFile(unittest.TestCase):
         path = self._write_json(devices)
         result = tinytuya.load_devicefile(path)
         self.assertEqual(result[0]['key'], ":|S'vf<MT6xhr{1~")
+
+
+class TestParseHeader(unittest.TestCase):
+    """parse_header must accept large IR/AC learn frames (>1000 bytes) while
+    still rejecting absurd sizes from a corrupt or desynced stream."""
+
+    def _header_6699(self, payload_len):
+        return struct.pack(
+            mh.H.MESSAGE_HEADER_FMT_6699,
+            mh.H.PREFIX_6699_VALUE, 0, 1, 8, payload_len,
+        )
+
+    def test_large_ir_payload_accepted(self):
+        # a 1038-byte frame is a real air-conditioner learn report; it must parse
+        header = mh.parse_header(self._header_6699(1038) + b'\x00' * 40)
+        self.assertEqual(header.length, 1038)
+
+    def test_oversized_payload_rejected(self):
+        oversized = self._header_6699(mh.MAX_PAYLOAD_LENGTH + 1)
+        with self.assertRaises(DecodeError):
+            mh.parse_header(oversized + b'\x00' * 40)
 
 
 if __name__ == '__main__':
--- a/tinytuya/core/const.py
+++ b/tinytuya/core/const.py
@@ -12,6 +12,12 @@ TIMEOUT = 3.0       # Seconds to wait for a broadcast
 TCPTIMEOUT = 0.4    # Seconds to wait for socket open for scanning
 DEFAULT_NETWORK = '192.168.0.0/24'
 
+# Heuristic ceiling used to reject corrupt/desynced streams. Most devices only
+# have ~256 KiB of RAM and need 2x-3x the payload size for buffers, plus packets
+# over ~1440 bytes tend to fragment, so keep the default conservative. Override
+# at runtime if a particular device needs larger frames.
+MAX_PAYLOAD_LENGTH = 1440
+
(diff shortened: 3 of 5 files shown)


uzlonewolf:
Thanks for the PR, @ostjen !

I feel 64 KiB is much too large.  Most of these devices only have 256 KiB of RAM total and they need to allocate 2x to 3x the payload size for all the buffers needed to base64 encode, JSON encode, and TCP packetize the data.  I also have doubts about how well they can cope with fragmented packets (over ~1440 bytes or so).

As such I would like to see the default reduced to 1440.  With the rest of this PR you could always increase that at runtime if your particular device needs something larger with something like:

```python
import tinytuya

tinytuya.MAX_PAYLOAD_LENGTH = whatever
```

The definition for `MAX_PAYLOAD_LENGTH` should also be moved into `const.py` instead of `message_helper.py`.

If you could make those 2 changes for me I can get this moving.

3735943886:
Changing the hardcoded `1000` bytes ceiling to a constant sounds like a great idea.

However, before adjusting the limit, there is a possibility that the learned IR code contains redundant or repeated signals. Sometimes, IR blasters capture the same frame multiple times if a button is held down, resulting in an unnecessarily large base64 payload—even if it technically works.

Please consider checking the long IR code with the snippet below.

```python
import tinytuya
from tinytuya import Contrib

# Paste your long IR base64 code here
ircode = "YOUR_LONG_IR_CODE_HERE"

pulses = Contrib.IRRemoteControlDevice.base64_to_pulses(ircode)
headkey = Contrib.IRRemoteControlDevice.pulses_to_head_key(pulses)

print("Parsed Head & Key:", headkey)

```

If the output shows a repetitive sequence, optimizing the IR code itself might be a cleaner solution than raising the protocol's sanity-check ceiling. What does the output look like for this code?

uzlonewolf:
I agree that de-duplicating the code is a good idea, however OP is unable to get the "YOUR_LONG_IR_CODE_HERE" due to the receive limit.

3735943886:
Good point, I assumed the OP got the long IR code from the cloud and ran into an error while trying to send it. But if the goal is to use full local control eg `receive_button()`, it's impossible to capture the code in the first place.
In that case, increasing the max value just to capture the raw data seems like a solid workaround, and optimizing the IR code is still worth looking into once it's obtained if the packet is too long.

ostjen:
Pushed the two changes in `0f38679` — default is now 1440 and lives in `const.py`. Thanks both!

@3735943886 since I could finally capture the full frames after bumping the limit locally, I ran your snippet on them. The head/key parses fine, and re-encoding `head_key_to_pulses → pulses_to_base64` gives back the exact same 936-char payload, so it's already in the minimal NCD form. The pulse train splits into three ~57-bit blocks on Samsung's ~8957µs lead spaces, but they're different bit patterns, not repeats — looks like the normal Samsung AC multi-block message rather than a held-button artifact.

The longest IR string I captured is 936 base64 chars; once that's wrapped in the JSON DP + Tuya envelope the full `payload_len` lands just over the old 1000-byte ceiling and well under 1440, which is what made this bump necessary.

jasonacox-sam (approved):
Clear fix for a real problem — IR learn frames from blasters like the wnykq Smart IR (protocol 3.5) can hit ~1038 bytes, which the old hardcoded `1000` ceiling in `parse_header()` rejected as corrupt.

**What I like about this:**
- Named constant (`MAX_PAYLOAD_LENGTH = 1440`) in `const.py` instead of a magic number — easy to find and adjust later
- 1440 is well-chosen: TCP MTU-aligned, respects device RAM constraints, and covers the reported 1038-byte frames with room to spare
- Clean import chain — `const.py` → `message_helper.py` → no unnecessary coupling
- Tests cover both `55AA` and `6699` prefix headers, valid large frames, and rejection of frames exceeding the new limit

Agree with @uzlonewolf that 64KiB would have been overkill — device-side buffers and TCP fragmentation make that impractical. 1440 is the right ceiling for now.

Nice first contribution, @ostjen. Ship-it from me, Jason's call on the merge.

— Sam ⚙️

jasonacox:
Thanks @ostjen @uzlonewolf and @3735943886 !

@jasonacox-sam please increment the library version and update the RELEASE.md in prep for a PyPI release.  Also, run a local test after that change.

jasonacox-sam (approved):
Clean, well-scoped fix. Replacing the magic `1000` with a named constant in `const.py` (defaulting to 1440) is the right call — generous enough for real IR frames, conservative enough to still catch desynced streams. Tests confirm both the accept and reject paths. Thanks for iterating on the ceiling per @uzlonewolf's feedback.

— Sam ⚙️

jasonacox-sam:
Done — verified the version bump and ran the full test suite:

- **Version:** `1.18.1` in `tinytuya/core/core.py` ✅
- **RELEASE.md:** Updated with v1.18.1 entry covering the `MAX_PAYLOAD_LENGTH` fix ✅
- **Tests:** All 22 tests pass, including `TestParseHeader` ✅
- **Rebased** on `upstream/master` — no divergence ✅

Ready for merge and PyPI publish whenever you are, @jasonacox.

— Sam ⚙️

jasonacox-sam:
Version bumped to `1.18.1` and `RELEASE.md` updated directly on this branch.

**Testing:** All 22 unit tests pass. Also ran live device testing on a real Tuya device (Kitchen Light, protocol 3.3) on the local network using the PR branch code — status check returned correctly. No regressions.

— Sam ⚙️

jasonacox:
View at:
https://pypi.org/project/tinytuya/1.18.1/

PR #711, opened 2026-06-02
Add FloorFanDevice support for Comfort Zone floor standing tower fan (CZTF423S)

cmoates:
- Implement FloorFanDevice class with support for:
  - Power control
  - Wind modes (normal, nature, sleep)
  - Speed levels (1-5)
  - Oscillation/swing control
  - Sleep timer (cancel, 1h-12h)
- Add comprehensive example usage in examples/Contrib/FloorFan-example.py
- Update Contrib __init__.py to export new device class

### DIFF
--- a/examples/Contrib/FloorFan-example.py
+++ b/examples/Contrib/FloorFan-example.py
@@ -30,8 +30,8 @@ def main():
     # Enable debug mode to see what's happening
     # fan.set_debug(True)
 
-    try:
-        print("=== Floor Standing Fan Control Example ===\n")
+    try:        # Note: Each status_json() call makes a full round-trip to the device.
+        # For production use, consider caching status or reducing polling frequency.        print("=== Floor Standing Fan Control Example ===\n")
 
         # Get the current status
         print("1. Getting current status...")
--- a/examples/Contrib/FloorFan-example.py
+++ b/examples/Contrib/FloorFan-example.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+"""
+Example usage of FloorFanDevice from tinytuya
+
+This example shows how to use the FloorFanDevice class to control
+a Tuya floor standing fan device locally.
+"""
+
+import sys
+from tinytuya.Contrib.FloorFanDevice import FloorFanDevice
+
+# Device Configuration
+DEVICE_ID = "YOUR_DEVICE_ID"
+IP_ADDRESS = "YOUR_DEVICE_IP"
+LOCAL_KEY = "YOUR_LOCAL_KEY"
+VERSION = 3.3
+
+def main():
+    """
+    Main example function demonstrating FloorFanDevice usage
+    """
+    # Initialize the device
+    fan = FloorFanDevice(
+        DEVICE_ID,
+        IP_ADDRESS,
+        LOCAL_KEY,
+        version=VERSION
+    )
+
+    # Enable debug mode to see what's happening
+    # fan.set_debug(True)
+
+    try:
+        print("=== Floor Standing Fan Control Example ===\n")
+
+        # Get the current status
+        print("1. Getting current status...")
+        status = fan.status_json()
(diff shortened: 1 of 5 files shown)


jasonacox-sam (commented):
Thanks for the contribution @cmoates! Clean Contrib module, follows the established patterns well.

A few observations for consideration:

1. **Silent failures on invalid input** — `set_mode()`, `set_speed()`, and `set_timer()` all silently `return` when given invalid values. Consider raising `ValueError` instead so callers know their command was rejected. Current behavior means a typo like `set_mode("natrue")` silently does nothing, which is hard to debug.

2. **DPS_STATUS_FLAG defined but unused** — the class defines `DPS_STATUS_FLAG = "13"` but it is not included in `status_json()` output and has no getter/setter. Is this DPS used by the device? If so, worth exposing. If not, consider removing the constant to avoid confusion.

3. **Missing `try/except` in status getters** — `get_power()`, `get_mode()`, etc. will raise `KeyError` if the DPS key is not present in the response (e.g., device offline returns empty `dps`). Other Contrib modules handle this gracefully — worth adding a `.get()` with a sensible default.

4. **Example uses real `status()` calls** — the example script calls `status_json()` after every command, which is great for documentation but note that each `status_json()` call is a full round-trip to the device. Users running the example should be aware of this.

Overall structure is solid — just the defensive coding patterns could be tightened up. No blocking issues from my side.

— Sam ⚙️

uzlonewolf:
Thanks @cmoates !

> Update Contrib \_\_init\_\_.py

As per the comments at the top of the file, this file is deprecated and new modules should not be added to it.

jasonacox:
Thanks @cmoates and thanks for updating per @uzlonewolf 

@jasonacox-sam please add a KB note for Contrib __init__.py for future reviews and Q&A:

Note: This file has been deprecated, please do not add new modules to it.
Instead, import new modules with `from tinytuya.Contrib import YourNewModule`
and call with `YourNewModule.YourNewModule(...)`

cmoates:
FWIW, it was testcontrib.py that got me into trouble there, I was trying to make sure any and all tests ran. Maybe break or remove that script, or something?

jasonacox:
@jasonacox-sam can you make sure testcontrib.py doesn't confuse others?

Let's include this PR on the new release.

jasonacox-sam:
@cmoates — you were absolutely right that `testcontrib.py` was confusing. It used the old `from tinytuya import Contrib` pattern that we've since deprecated, so anyone reading it would naturally follow that example. That's exactly what happened here, and it's on us for not updating it sooner. Sorry about that!

I've pushed two updates to this PR branch:

**1. `testcontrib.py` — updated to the preferred import pattern**
- Added a clear deprecation header at the top explaining that `Contrib/__init__.py` is deprecated and new modules should use `from tinytuya.Contrib import YourNewModule`
- Switched all device instantiations to use direct imports (`ThermostatDevice(...)`, `InverterHeatPumpDevice(...)`, `TemperatureUnit(...)`) instead of `Contrib.ThermostatDevice(...)`
- Kept the backward-compatible `DeviceTypes` listing (still useful for visibility) but moved it below the preferred imports with a comment marking it as legacy

**2. `tinytuya/Contrib/README.md` — consistent examples**
- Updated the `ThermostatDevice`, `IRRemoteControlDevice`, and `InverterHeatPumpDevice` examples to all use the direct import pattern — some already did (SocketDevice, DoorbellDevice, etc.), but these three were still showing the old way

Your `FloorFanDevice` module itself looks great — clean DPS mapping, good docstrings, proper error handling on `set_mode`/`set_speed`/`set_timer`. Thanks for the contribution!

@jasonacox — both items addressed. This PR is ready for the next release. 🚀

— Sam ⚙️

jasonacox:
Thanks @cmoates ! 🙏  This will be included in our next release.

PR #714, opened 2026-06-13
fix: call detect_bulb() in *_percentage() methods before reading value_max

### LINKED ISSUE #713: set_brightness_percentage() - not working properly
Hello,
Since few days, if I set brightness with set_brightness_percentage(val) is set-up every time to 100%.
If I set-up using set_brightness(val) between 0-1000 all is good.
The bulb is LEDVANCE and since 2021 I didn't have other problems.

import tinytuya
d = tinytuya.BulbDevice('device_id', '192.168.10.60','local_key')
d.set_version(3.3)
d.set_brightness_percentage(75)
print(d.status())

and output: {'dps': {'20': True, '21': 'white', '22': 1000, '23': 488, '24': '006103330334', '25': '000e0d0000000000000000c80000', '26': 0, '101': False}}

Is this a bug or my bulb has a problem?

Thank you!
Andrei

jasonacox-sam:
Fixes #713

## Problem

When `set_brightness_percentage()`, `set_white_percentage()`, or `set_colourtemp_percentage()` is called **before bulb detection has run**, `self.dpset['value_max']` is still `-1` (the unconfigured default). The computed raw value is wrong, which causes the `< 0` fallback in `set_brightness()` to fire and pin the device to full brightness regardless of the requested percentage.

## Fix

Add a `if not self.bulb_configured: self.detect_bulb(nowait=nowait)` guard at the top of each `_percentage()` method, before the `value_max` multiplication. This mirrors the guard already present in `set_brightness()`, `set_white()`, and other lower-level methods.

## Affected Methods

- `set_brightness_percentage()`
- `set_white_percentage()`
- `set_colourtemp_percentage()`

## Testing

Workaround confirmed by @andreidiaconescu18 in #713 — calling `d.status()` first (which triggers `detect_bulb()`) makes `set_brightness_percentage(75)` work correctly. This PR makes that detection automatic.

— Sam ⚙️

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -259,6 +259,22 @@ class TestXenonDevice(unittest.TestCase):
         self.assertEqual(result_cmd, expected_cmd)
         self.assertDictEqual(result_payload, expected_payload)
 
+    def test_percentage_raises_when_unconfigured(self):
+        # Regression test: calling *_percentage() before bulb detection,
+        # with nowait=True and no cached status, should raise RuntimeError
+        # instead of silently computing with value_max=-1.
+        d = tinytuya.BulbDevice('DEVICE_ID_HERE', 'IP_ADDRESS_HERE', LOCAL_KEY)
+        d.set_version(3.3)
+        # No detect_bulb() called, no cached status available
+        d.cached_status = MagicMock(return_value=None)
+
+        with self.assertRaises(RuntimeError):
+            d.set_brightness_percentage(50, nowait=True)
+        with self.assertRaises(RuntimeError):
+            d.set_white_percentage(50, 50, nowait=True)
+        with self.assertRaises(RuntimeError):
+            d.set_colourtemp_percentage(50, nowait=True)
+
 
 def build_mock_rf():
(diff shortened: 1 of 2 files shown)


jasonacox:
@jasonacox-sam Please review and address the copilot code review items if they have merit.

jasonacox-sam:
@jasonacox — Reviewed all four Copilot items. They have merit: `detect_bulb()` can exit without configuring when `nowait=True` and there's no cached status, leaving `value_max` at its default of `-1`. The existing `state()` and `bulb_has_capability()` methods already guard against this with a post-detect `bulb_configured` check — the `*_percentage()` methods were missing the same guard.

Fixed in 3841d1b:
- Added post-detect `RuntimeError` guard to `set_brightness_percentage()`, `set_white_percentage()`, and `set_colourtemp_percentage()` — matching the pattern already used by `state()` and `bulb_has_capability()`
- Added `test_percentage_raises_when_unconfigured` regression test that verifies all three methods raise `RuntimeError` when called before detection with `nowait=True` and no cached status

All 23 tests passing. — Sam ⚙️

jasonacox:
Thanks, Sam! Added update to bundle with future release.

PR #715, opened 2026-06-25
Fix for "Error in main loop: 'id'" of Force Scan

mkerni:
Force-scanned devices are returned keyed by IP and are not guaranteed to carry an "id" (only set when the local key is matched). Fall back to "gwId" and skip entries without one to avoid raising KeyError: 'id' in the main loop.

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -759,7 +759,10 @@ if __name__ == "__main__":
                 print(f" - ForceScan: Found {len(found)} devices")
                 for f in found:
                     log.debug(f"   - {found[f]}")
-                    gwId = found[f]["id"]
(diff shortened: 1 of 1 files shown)


jasonacox-sam (approved):
Good catch — this is a real crash. ForceScan results are keyed by IP, and entries aren'''t guaranteed to have an "id" until a local key match occurs. Falling back to "gwId" and skipping entries with neither prevents the KeyError without losing devices that do have identifying info.

The debug log for skipped entries is a nice touch — makes the skip visible during troubleshooting instead of silently ignoring.

LGTM. — Sam ⚙️

jasonacox (commented):
Good addition.

PR #717, opened 2026-06-25
Cloud: send PUT/DELETE with the real HTTP method (fixes #716)

### LINKED ISSUE #716: Cloud._tuyaplatform sends PUT/DELETE as POST → 1108 'uri path invalid'
## Summary

`Cloud._tuyaplatform()` computes the request signature using the real HTTP method (e.g. `PUT`/`DELETE`), but the request dispatch only special-cases `GET` and sends **everything else via `requests.post()`**:

```python
if action == 'GET':
    response = requests.get(url, headers=headers)
else:
    response = requests.post(url, headers=headers, data=body)
```

So any endpoint that requires `PUT` or `DELETE` actually receives a `POST`. Tuya then rejects it with `{"code": 1108, "msg": "uri path invalid", "success": false}` because the path only exists for the correct verb. The mismatch is easy to miss because the signature header is valid (it was built with the intended method).

## Reproduce

Rename a device (docs: *Modify device name*, `PUT /v1.0/devices/{device_id}`):

```python
import tinytuya
c = tinytuya.Cloud(apiRegion='eu', apiKey='...', apiSecret='...')
print(c.cloudrequest('/v1.0/devices/<id>/..', action='PUT', post={'name': 'New name'}))
# -> {'code': 1108, 'msg': 'uri path invalid', 'success': False}
```

`GET` calls on the same project work fine, and the `Device Maintenance` API is authorized — only `PUT`/`DELETE` fail.

## Fix

Dispatch with the same method that was signed:

```python
if action == 'GET':
    response = requests.get(url, headers=headers)
else:
    response = requests.request(action, url, headers=headers, data=body)
```

With this change the rename `PUT` returns `{'result': true, 'success': true}`.

Affects current `master` and v1.18.1. PR with the one-line fix to follow.

vladulus:
Fixes #716.

`Cloud._tuyaplatform()` signs each request with the actual action (`PUT`/`DELETE`/…) but dispatches every non-`GET` call through `requests.post()`. Endpoints that require `PUT` or `DELETE` therefore receive a `POST` and Tuya rejects them with `1108 'uri path invalid'` — most visibly the device-rename endpoint `PUT /v1.0/devices/{device_id}`.

This swaps the catch-all `requests.post()` for `requests.request(action, ...)`, so the dispatched method matches the one already used to build the signature. `GET` keeps its dedicated branch; `POST` behaviour is unchanged.

Verified against the live EU cloud: before, the rename `PUT` returns `1108`; after, it returns `{'result': true, 'success': true}` and the new name shows up in the Smart Life app.

Co-authored-by: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/tinytuya/Cloud.py
+++ b/tinytuya/Cloud.py
@@ -231,11 +231,13 @@ class Cloud(object):
             )
         else:
             log.debug(
-                "POST: URL=%s HEADERS=%s DATA=%s", url, headers, body,
+                "%s: URL=%s HEADERS=%s DATA=%s", action, url, headers, body,
             )
-            response = requests.post(url, headers=headers, data=body)
+            # use the actual HTTP method (the signature is already computed with it);
+            # always sending POST here breaks PUT/DELETE endpoints (e.g. device rename)
+            response = requests.request(action, url, headers=headers, data=body)
             log.debug(
-                "POST RESPONSE: code=%d text=%s token=%s", response.status_code, response.text, self.token
+                "%s RESPONSE: code=%d text=%s token=%s", action, response.status_code, response.text, self.token
             )
 
         # Check to see if token is expired


jasonacox-sam (approved):
Nice catch — this is a real bug. The signature at line 203 already includes `action` as the HTTP method in the signing string, but the dispatch was hardcoded to `requests.post()` for everything non-GET. That means PUT/DELETE endpoints were signed correctly but sent as POST, so Tuya rejects them with `1108`.

Using `requests.request(action, ...)` makes the dispatched method match the signed method. The debug log update is a good touch too — `'POST RESPONSE'` was misleading for PUT/DELETE calls.

LGTM — minimal change, correct fix. Thanks @vladulus!

— Sam ⚙️

jasonacox:
Thanks @vladulus 🙏

PR #735, opened 2026-09-03
Scanner: poll configured static-IP devices that do not broadcast

merodahero:
## Summary

- poll fully configured static-IPv4 devices when UDP discovery does not identify them
- reuse TinyTuya's existing `DeviceDetect` / `PollDevice` / `select()` state machine, bounded by `max_parallel` and ten configured starts per event-loop pass
- keep UDP authoritative, including late announcements for the same device ID at a different IP
- validate candidates before socket construction and keep malformed rows or failed polls non-fatal
- reconcile static results, verbose totals, version counts, and snapshots without duplicates or local-key output

## Problem

Some Tuya devices accept authenticated TCP/6668 connections but do not send a usable UDP discovery response. TinyTuya can therefore have every value needed to poll the device in `devices.json` and still omit it from an ordinary scan.

This behavior was observed with a protocol 3.5 breaker/meter. A direct `OutletDevice.status()` call succeeds, while UDP and forced discovery do not identify it. No device-specific values are included in this change.

This follows the non-broadcasting-device discussion in #511, particularly the concern that silent devices might become a broader firmware trend and the suggestion to poll known IPs directly.

## Approach

UDP discovery and explicit force scanning run first. Afterwards, device-list rows are eligible only when they contain a non-empty ID and local key, a valid IPv4 address, and a supported protocol version (3.1-3.5).

Eligible devices enter the existing non-blocking scanner lifecycle in bounded batches. A late UDP announcement supersedes a pending, in-flight, or completed static result by device ID or IP. Matching static sockets/results are closed and removed before result shaping. Failed static polls are omitted without aborting the scan, and remaining sockets are cleaned up on normal finalisation or interruption.

Successful fallback results carry `origin = "static"` and participate once in verbose totals, protocol-version counts, return data, and snapshot output. Verbose static output uses a redacted copy of device metadata.

## Tests

Verified locally:

- `python3 -m unittest tests.TestConfiguredDeviceScanner -v` - 15 passed
- `python3 -m tests` - 53 run, 52 passed, 1 skipped
- `python3 -m compileall -q tinytuya` - passed
- `python3 testcontrib.py` - passed
- `pylint --recursive y -E tinytuya/` - passed
- `git diff --check` - passed
- Python 3.7 container full suite - 53 run, 52 passed, 1 skipped
- Python 3.12 container full suite - 53 run, 52 passed, 1 skipped
- sabotage check disabling configured-candidate selection - central regression failed as expected, then passed after restoration

Regression coverage includes candidate validation, alternate device files, bounded later-round scheduling, failed connects/polls, same-ID late UDP at a different IP, completed-result invalidation, interruption cleanup, verbose reconciliation/key secrecy, and a real `StaticDevice` connect/write/read lifecycle with only socket/wire primitives mocked.

## Hardware validation

Validated the final branch source against a real protocol 3.5 device that:

- has TCP/6668 open;
- returns live DPS through the configured static poll;
- does not provide a usable UDP discovery response;
- appears once as `[Static Poll]` with `origin = "static"`;
- is counted once in device and protocol-version totals;
- does not expose its local key in captured output.

Private device identifiers, addresses, MACs, keys, names, and DPS values are intentionally omitted.

## Risk

The change is limited to scanner discovery/poll orchestration, adds no dependency, and reuses TinyTuya's authenticated read-only polling path. Only explicitly configured, strictly validated devices are contacted, after the ordinary discovery window, with bounded concurrency.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -21,6 +21,7 @@ are listed here so the release changelog is complete.
 
 ### Bug fixes (see bug-fix PR)
 * Cloud token-refresh retry preserves the query string; Cloud helpers return error dicts instead of raising on failed/empty responses.
+* Scanner: Fully configured `devices.json` entries (ID, local key, IPv4 address, and protocol version) are now polled in bounded parallel batches after UDP discovery, allowing non-broadcasting devices to be found while late broadcasts remain authoritative and duplicate-free. PR #TBD by @merodahero.
 * `set_timer()` selects the timer DP numerically; timeout errors report `ERR_TIMEOUT`; `received_wrong_cid_queue` is bounded; truncated frames raise `DecodeError`; `error_json()` handles unknown codes; `BulbDevice` rgb8 hue offset fixed; scanner no longer mutates caller dicts; and more.
 
 ## v1.19.0 - Monitor Class, IPv6, and Community Fixes
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -21,7 +21,7 @@ are listed here so the release changelog is complete.
 
 ### Bug fixes (see bug-fix PR)
 * Cloud token-refresh retry preserves the query string; Cloud helpers return error dicts instead of raising on failed/empty responses.
-* Scanner: Fully configured `devices.json` entries (ID, local key, IPv4 address, and protocol version) are now polled in bounded parallel batches after UDP discovery, allowing non-broadcasting devices to be found while late broadcasts remain authoritative and duplicate-free. PR #TBD by @merodahero.
+* Scanner: Fully configured `devices.json` entries (ID, local key, IPv4 address, and protocol version) are now polled in bounded parallel batches after UDP discovery, allowing non-broadcasting devices to be found while late broadcasts remain authoritative and duplicate-free. PR #735 by @merodahero.
 * `set_timer()` selects the timer DP numerically; timeout errors report `ERR_TIMEOUT`; `received_wrong_cid_queue` is bounded; truncated frames raise `DecodeError`; `error_json()` handles unknown codes; `BulbDevice` rgb8 hue offset fixed; scanner no longer mutates caller dicts; and more.
 
 ## v1.19.0 - Monitor Class, IPv6, and Community Fixes
--- a/tests.py
+++ b/tests.py
@@ -4,9 +4,9 @@ import logging
 
 import unittest
 try:
-    from unittest.mock import MagicMock  # Python 3
+    from unittest.mock import MagicMock, patch  # Python 3
 except ImportError:
-    from mock import MagicMock  # py2 use https://pypi.python.org/pypi/mock
+    from mock import MagicMock, patch  # py2 use https://pypi.python.org/pypi/mock
 import json
 import struct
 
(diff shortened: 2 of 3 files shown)


jasonacox-sam (commented):
This is a really thorough piece of work — the test suite in particular (bounded-batch scheduling, late-broadcast invalidation, interruption cleanup, key redaction in verbose output) covers exactly the corners I'd worry about. Nice job.

A few design questions before this is ready:

**1. Scan-duration cost with stale entries (main concern).**
With `poll=True`, this now contacts every fully-configured row in `devices.json` on *every* scan, after the UDP window closes. For households where IPs rotate via DHCP, `devices.json` tends to accumulate stale addresses — and each stale entry costs a full TCP connect timeout inside the scan, every single time. On a 20-device list with a handful of dead entries, `tinytuya scan` gets visibly longer forever. Two thoughts:
- Consider an opt-out parameter (e.g. `devices(poll_configured=...)`) or a way for CLI users to skip it, and/or
- Only poll configured entries that look unreachable by UDP after N consecutive scans — though that adds state, so a simple opt-out may be enough.

**2. Hardcoded version membership.**
`version not in (3.1, 3.2, 3.3, 3.4, 3.5)` silently skips future protocol versions (3.6+). That's safe-by-default, which I like, but it will confuse someone whose 3.6 device is skipped with no output. A `log.debug` on the skip (or surfacing it in verbose mode) would save a future issue report.

**3. Minor: float equality on versions.**
`float(item['version'])` handles string versions like `"3.3"` fine, and tuple membership is exact for these values — no bug, just noting it's fragile if the supported-version list ever gets derived dynamically.

The UDP-authoritative reconciliation (including same-ID-at-different-IP supersede) and the `StaticDevice.close()` "only a decoded status response counts as found" rule are both the right calls. — Sam ⚙️

merodahero:
Thanks for the close read, Sam — those are great questions.

**1. Stale-entry cost:** agreed. I’ll take the simple stateless path and add a compatibility-preserving opt-out: `poll_configured=True` as a final API parameter, plus `scan -no-poll-configured` for CLI users. The default will retain the fallback introduced here; opting out will leave ordinary UDP/forced discovery intact while skipping configured static-IP polls. I won’t add scan-history state or mutate `devices.json`.

**2. Future versions:** agreed. I’ll emit a debug message when an otherwise complete configured row is skipped for an unsupported protocol version, without including its local key.

**3. Float equality:** noted. TinyTuya’s device APIs currently use float protocol versions, so I’ll keep that representation but centralise the pinned supported-version tuple and document that exact membership is intentional for the literal 3.1–3.5 set.

I’ll add focused regression coverage and update the public API/CLI documentation in the same follow-up commit.

jasonacox-sam:
That plan covers all three points cleanly — the stateless opt-out is the right call (no hidden scan-history state, no surprise `devices.json` mutation), and skipping the key in the debug message is exactly the right instinct. Looking forward to the follow-up commit.

— Sam ⚙️

jasonacox-sam:
Went through Copilot's latest pass against the branch — I agree with most of it, with one scoping note. In rough priority order:

**Agreed, worth fixing in this PR:**

1. **Early-termination gap** — `ip_force_wants_end` isn't checked in the static-poll launch condition. Once wanted IPs/IDs are found (or `maxdevices` is hit), `scan_end_time` is zeroed, which makes `scan_end_time <= time.time()` immediately true — so stale configured addresses can still be contacted after the caller asked the scan to stop. Adding `not ip_force_wants_end` to the gate looks like the clean fix.

2. **`maxdevices` not enforced on the static path** — the limit is only decremented in the UDP-discovery branch, so `devices(maxdevices=1, poll_configured=True)` with no broadcasts can return every configured device. Carrying the remaining count through static result acceptance seems right.

3. **Force-scan drain race** — when the force-scan generator is exhausted, `ip_scan_running` goes false while in-flight `ForceScannedDevice` objects are still in `devicelist`, and the duplicate check only inspects the broadcast/scanned/configured maps — so a `StaticDevice` can open on the same IP and be counted separately. Including active devices' IPs/IDs in the eligibility check (or draining first) resolves it.

4. **`abort()` retry race** — if a UDP socket and a static socket are both ready in one `select()` pass and the UDP branch aborts the static poll, the later read branch calls `read_data()` with `self.sock` already `None`, and its exception path can retry/reconnect, undoing the abort. Skipping sockets that were closed/removed during dispatch (or making an aborted poll non-retryable) both work.

**Scoping note — pre-existing, not introduced here:** the truthy-only checks in the `keylist` loop and the verbose snapshot loop's `item["id"]` indexing are identical to upstream `master` (they predate this PR). Hardening them is worthwhile, but I'd treat that as a separate follow-up rather than expanding this PR's diff — malformed devices.json rows already crash verbose scans on `master` today.

— Sam ⚙️

merodahero:
Addressed the scanner review findings in b46f00fdf9c41f1689ba5d6f76c91567a779807c, with regression tests and documentation updates:

- Validate ID/key types before static candidate and key-list construction; skip malformed rows in snapshot assembly and scanner lookups.
- Make static cancellation non-retryable and ignore stale `select()` events for closed/replaced sockets.
- Stop configured scheduling after early termination, reserve `maxdevices` capacity for in-flight polls, and release slots after failed polls. Broadcast and force-scan results consume the same fallback budget.
- Wait for active force scans to drain and register newly completed results before static duplicate checks.
- Reconcile requested IDs/IPs during polling and final reporting. Explicit force scans still run first; configured polling precedes the implicit force scan of missing requested IPs.
- Add `finally` cleanup for tracked scanner sockets, including connections opened since the previous `select()` pass. Cover interrupts during construction, connection, and the first `select()` call.
- Keep `snapshot()` and `snapshotjson()` scoped to saved snapshot addresses by explicitly disabling the configured fallback.

The default-enabled fallback and API/CLI opt-out remain unchanged.

**Local verification (Python 3.12):** 37 focused scanner tests passed; the full suite ran 75 tests with one skipped and no failures. Contrib checks, compileall, errors-only pylint, and `git diff --check` passed. Python 3.7 syntax compatibility was also checked (not a 3.7 runtime test).

**Hardware verification:** ran the exact patched source from an isolated temporary directory against a non-broadcasting protocol-3.5 meter. With the fallback disabled, discovery returned five broadcast devices and omitted the meter. With it enabled, discovery returned those five plus the meter as `origin=static` with 17 DPS fields. Direct status succeeded before and after. The installed package, device/snapshot files, and breaker state were unchanged; temporary source cleanup was verified.

The new upstream Actions runs currently report `action_required`; local and hardware results above are separate from upstream CI.

